PR #315, opened 2013-04-12
Moved assets.js() to footer

dylanized:
Moved Javascript asset block to footer, for better performance. Now scripts will be loaded after the page renders.

Might be worth having both a headerjs and footerjs folder, so assets can be bundled and inserted into either place?
- This is my 1st pull request ever, apologies for any fumbles!


jasalazar:
Hello, everyone. I know that for best practices all scripts should be placed at the end of the Body, but i am having troubles because of this.

Is there any place where i can change this configuration to place my .js files located @ assets folder in the headers?

Thanks a lot!

mikermcneil:
@jasalazar if I'm understanding you right, you should actually just be able to move the `<!-- -->` comments up to where you want the scripts to be injected, and grunt-sails-linker will insert them in the appropriate place.  Hope that helps!

jasalazar:
That absolutely worked, @mikermcneil thanks a lot!

I'm loving and enjoying the Sails way!

PR #1441, opened 2014-02-24
Allow controller to override its `req.model`.

marionebl:
Allows controllers to override the model id passed into `req.model`:

``` javascript
// api/controllers/ThisController.js
module.exports = {
  _config: {
    model: 'that'
  }
}
```

Would enable blueprint routes for `/this/*`, using a hypothetical model `That`.

Most prominent use case would be a short-term solution for namespaced controllers missing out on all blueprint functionality as there is no possibility to namespace models at the moment.

### DIFF
--- a/lib/hooks/blueprints/index.js
+++ b/lib/hooks/blueprints/index.js
@@ -216,9 +216,11 @@ module.exports = function(sails) {
 				}
 			});
 
-      // Get the model connected to this controller either by explicit configuration
-      // or implicit by controller id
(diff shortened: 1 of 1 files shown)


mikermcneil:
@marionebl So it's been like, 8 years, but I was thinking recently about some of the stuff we discussed way back when.  Want to catch up?  (sent you a dm on LinkedIn, https://www.linkedin.com/in/mikermcneil/)

PR #1590, opened 2014-04-12
Option to set the populate settings from the association itself

Skullbock:
With this change, you can set the populate settings of an association directly in the association object itself.
This allows the developer to continue using the blueprint for REST apis even when the association has to be limited, sorted or filtered in a different way that the standard one


CWyrtzen:
Bumped~

tjwebb:
@Skullbock can you merge in latest into your PR? I'd like to merge this.

mikermcneil:
@Skullbock @tjwebb I'm going to go ahead and close this- if you're still interested in this, please update it to the latest master, submit a new PR, and I'll merge it in.  Thanks!

Issue #1598, opened 2014-04-14
Improve environment settings


Kikobeats:
Hi,

I'm using v0.10 and I have checked that change from dev environment to production or vice versa is complicated because the code is very disperse.

I can to propose a better way to organize inspired in  [Yoi Server](https://github.com/tapquo/yoi):
- create folder `config/env` and create differents files for differents environments, for example `config/env/development` and `config/env/production`.
- rename `config/local` file  to `config/environment` and use this file only for select what environment from `config/env` select
- merge sensible and necessary code from old `config/local` and `config/connections` in environments files

Now, the code dependently sensitive of the environment is concentred in a same file, and change environment is very easy changing one word in one file.

What do you think?

PR #1638, opened 2014-04-29
Improve environment settings inspired by #1598

clark0x:
Add an optional `env` dir under `config`. If a config file for the current environment exists, it will be loaded and override `local.js`.

P.S.: I think `excludeDirs` option for `include-all` only accepts RegExp, `Array` doesn't work.


laurynas-karvelis:
Hi all,

first of all, sails.js is an awesome piece of work.

I've just noticed that on Sails.js 0.11.0 final version config/env/{environment}.js does not override default /config/*.js configuration if given config attribute is an array.

Could this be fixed?
Thank you so so much in advance!

Issue #1932, opened 2014-07-08
Policy doesn't protected blueprint action with custom route option


clarkorz:
$ sails generate new sails-rc8
    $ sails generate api user

    // edit config/policies.js
    module.exports.policies = {
	  '*': false
	}

     // edit config/routes.js
    module.exports.routes = {
	  'GET /user': { model: 'user', blueprint: 'find', populate: false }
	}

    $sails lift
    $ curl http://127.0.0.1:1337/user

### Expect

Response 403 Forbidden

### Actual

Response 200

loicsaintroch:
I close this issue in favor of your PR. Thank you for your contribution @clarkorz, we appreciate. I'll take a look at the PR as soon as I have more time, sorry for the delay.

PR #1934, opened 2014-07-08
fix policy doesn't protected blueprint action in some cases

clarkorz:
1. fix policy doesn't protected blueprint action with custom route option
2. fix policy doesn't protected custom blueprint action
3. fix sails fails to lift if using in-exist model in route config

https://github.com/balderdashy/sails/issues/1932

### DIFF
--- a/lib/hooks/blueprints/onRoute.js
+++ b/lib/hooks/blueprints/onRoute.js
@@ -98,6 +98,12 @@ module.exports = function (sails) {
         );
         return;
       }
+    } else if (sails.models && !sails.models[options.model]) {
+      sails.log.error(
+        blueprintActionID,
+          ':: Ignoring attempt to bind route (' + path + ') to unknown model (`'+options.model+'`).'
+      );
+      return;
     }
 
     // If associations weren't provided with the options, try and get them
@@ -131,7 +137,7 @@ module.exports = function (sails) {
       options.populate = sails.config.blueprints.populate;
     }
 
-		sails.router.bind(path, blueprint, verb, options);
+		sails.router.bind(path, { controller: options.model, action: blueprintActionID }, verb, options);
 
 		return;
 	}
--- a/lib/hooks/controllers/onRoute.js
+++ b/lib/hooks/controllers/onRoute.js
@@ -214,7 +214,7 @@ module.exports = function (sails) {
 			return;
 		}
 
-		sails.router.bind(path, blueprint, verb, options);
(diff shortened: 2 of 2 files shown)


CWyrtzen:
Bumped~

sgress454:
Ok, I'll will take another look at this--I think this is probably a question of making sure our tests match our [documented expectations](http://sailsjs.org/#!/documentation/concepts/Routes/RouteTargetSyntax.html).

kevinob11:
Any word on this? I just ran into this today.

Bazze:
This is interesting. I ran in to the same thing. After several attempts I don't seem to be able to apply policies to a custom route. Status on this?

For now you can solve it with a really ugly workaround, as long as you only need to apply a few policies (otherwise it will get a complete mess). I call my policy with my custom blueprint function as the `next` callback function, like this:

```javascript
var authTokenPolicy = require('../policies/authToken');

module.exports = function (req, res) {
  authTokenPolicy(req, res, function getCount() {
    var Model = actionUtil.parseModel(req);
    var criteria = actionUtil.parseCriteria(req);

    Model
      .count(criteria)
      .then(function (response) {
        res.ok({count: response});
      })
      .catch(function (error) {
        return res.serverError('database_error', error);
      });
  });
};
```

sgress454:
@kevinob11 Policies do protect the blueprints that come with Sails.  This issue refers to _custom_ blueprints, and custom routes that use the [blueprint target syntax](http://sailsjs.org/documentation/concepts/routes/custom-routes#?blueprint-target-syntax).

@Bazze you can always add policy protection to a custom route using the [policy target syntax](http://sailsjs.org/documentation/concepts/routes/custom-routes#?policy-target-syntax) in `config/routes.js`.  We'd like to have the `config/policies.js` file cover those as well, hence this issue.

Bazze:
@sgress454 Thank you for that link! Much better workaround.

Bazze:
@sgress454 Do you know if there is a way to specify the model for the policy target syntax? It doesn't seem to work as with the blueprint target syntax. I'm doing an `actionUtil.parseModel()` in my policy which results in an error: `No "model" specified in route options.`

sgress454:
@Bazze this is [fixed in the latest code](https://github.com/balderdashy/sails/commit/0990fc10709520a9f6c55923b991708d5eaf8aa0), but the patch hasn't been released yet.  You can either apply the patch yourself or check out the latest code from Github.

PR #1972, opened 2014-07-18
[Don't merge] - Added criteria support on associations for blueprints

RWOverdijk:
This PR is my first stab at allowing the usage of filters in the url. I didn't take a lot of time writing it so it's probably flawed.

Let the discussion start :)


mikermcneil:
> I'd say that the populate array only dictates which models to populate.
> 
> Now that custom blueprints can be defined, and you can protect them with policies, too, I think this issue has less of a priority.

@RWOverdijk ok roger that.  I'm going to close this to keep tidy but let's keep the conversation going

Issue #1268, opened 2014-01-03
FEATURE_REQUEST: Add PATCH blueprint route


ragulka:
I couldn't find information on this - is there support for PATCH verb/method yet? If not, should we perhaps go ahead and start working on it?

It would be very useful when doing small, atomic updates to multiple objects, as described here: http://stackoverflow.com/questions/11872923/rest-updating-multiple-resources

PR #2006, opened 2014-07-23
Send PATCH requests to the "update" blueprint.

ProlificLab:
It makes it possible to deal with PATCH requests in blueprints, which is important to use [JSON PATCH](http://tools.ietf.org/html/rfc6902) and [JSON API](http://jsonapi.org/). It is linked to feature request #1268.

### DIFF
--- a/lib/hooks/blueprints/index.js
+++ b/lib/hooks/blueprints/index.js
@@ -312,6 +312,7 @@ module.exports = function(sails) {
             _bindRoute(_getRoute('post %s'), 'create');
             _bindRoute(_getRoute('put %s/:id'), 'update');
             _bindRoute(_getRoute('post %s/:id'), 'update');
+            _bindRoute(_getRoute('patch %s/:id'), 'update');
             _bindRoute(_getRoute('delete %s/:id?'), 'destroy');
 
             // Bind "rest" blueprint/shadow routes based on known associations in our model's schema
--- a/lib/hooks/cors/index.js
+++ b/lib/hooks/cors/index.js
@@ -20,7 +20,7 @@ module.exports = function(sails) {
 			cors: {
 				origin: '*',
 				credentials: true,
-				methods: 'GET, POST, PUT, DELETE, OPTIONS, HEAD',
+				methods: 'GET, POST, PUT, PATCH, DELETE, OPTIONS, HEAD',
 				headers: 'content-type'
 			}
 		},


CWyrtzen:
Bumped~

MarkMurphy:
I can probably duplicate the test @sgress454 linked to above and tweak it as required for testing `PATCH`. Shouldn't take long to do that. Might give it a go tonight.

CWyrtzen:
thanks @MarkMurphy~

MarkMurphy:
Can someone who knows the project well please resolve the merge conflicts before I write any tests for this.

tjwebb:
@loicsaintroch mind taking another look at this? re: https://github.com/balderdashy/sails/pull/2006#issuecomment-83544203

mikermcneil:
@tjwebb @loicsaintroch @MarkMurphy this changes the API for blueprints-- which is ok, albeit potentially confusing. To clarify, you can use `PATCH` with blueprint actions now-- it's just not automatic via shadow routes. For example, if you have blueprints enabled, and an empty model and controller called `Foo` and `FooController`, you can achieve the goal of this pull request in your own project by adding this route:

```js
'patch /foo/:id':  'FooController.update'
```

While I understand the interest in having PATCH shadow routes exist automatically, it's potentially confusing since the expectation could easily be that the behavior is different than using PUT.  I think it'd be a better step for the framework if we focused PRs like these on enhancing configurability in the blueprint hook without changing the default behavior.

For instance, we could expose a new, lower-level configuration property: `sails.config.blueprints.shadowRoutes`

```js
shadowRoutes: [
  {
    method: 'patch',
    address: '/:id',
    action: 'update'
  },
  {
    method: 'put',
    address: '/:id',
    action: 'update'
  },
  {
    method: 'post',
    address: '/',
    action: 'create'
  },
  {
    method: 'get',
    address: '/:id',
    action: 'findOne'
  },
  {
    method: 'get',
    address: '/',
    action: 'find'
  }
]
```

But in general, the best thing we can do here is pull out the blueprints hook into a separate module like I did with sockets for v0.11.  This will make it much easier for everyday users to use custom forks of the blueprints hook while still keeping up to date with the rest of Sails core.  You can technically already do this, but without a platonic example of the blueprints hook existing standalone, it's pretty intimidating for devs who haven't delved into the rest of Sails core.

Let's leave this PR open for now-- if you're interested in leading the charge with either of the two things I mentioned above, please let me know.

MarkMurphy:
@mikermcneil I totally agree with you and that's actually a really cool idea.

> it's potentially confusing since the expectation could easily be that the behavior is different than using PUT.

I was actually assuming `PUT` would be used to replace the resource not update it so your point about that is well received.

:+1:

dominikkukacka:
Yeah, this is should be expected behaviour:
  * PUT - exchanges the data on the server
  * PATCH - updates the specified fields (if they exist) or creates them but leaves everything else untouched

tjwebb:
@mikermcneil @sgress454 @particlebanana what are your thoughts on trying to get this into 0.12 GA?

listepo:
+1

MarkMurphy:
:zzz:

mikermcneil:
@MarkMurphy :sleeping: :deciduous_tree: :zap: 

@listepo @dominikkukacka @listepo @loicsaintroch @ProlificLab If any of you guys are interested in the hypothetical `sails.config.blueprints.shadowRoutes` config I started speccing out above, it'd be awesome if you could [put together a proposal](https://github.com/balderdashy/sails/blob/master/CONTRIBUTING.md#v-proposing-features-and-enhancements) extending/tweaking the stuff above, sharing your use case, and putting together a reference of all the files in the core blueprint hook, tests, generators, and documentation pages that would need to be updated.  If not, no problem-- let's go ahead and close this one up.  Thanks everyone for their feedback.

re: pulling out the blueprints hook as a separate package like https://github.com/balderdashy/sails-hook-sockets and https://github.com/balderdashy/sails-hook-orm, I'm planning on taking care of this in Sails v1 (makes it easier for the community to make packages like https://github.com/mikermcneil/sails-hook-orm-mongoose, and e.g. customize the blueprints hook for ember, etc.). More on that [in this proposal](https://github.com/balderdashy/sails/pull/3504).



   if you're interested in leading the charge with either of the two things I mentioned above, please let me know.

mikermcneil:
Guys, I'm going to go ahead and close this.  If anyone has interest in working on this, please open a proposal based on what we discussed above and link back to this PR for context.  Thanks!

PR #2053, opened 2014-08-04
Fixed the problem where parseSkip and parseLimit don't cover the case we...

weyert:
Fixed the problem where parseSkip and parseLimit don't cover the case were req.options.limit/skip returns `null` instead of `undefined`. This fixes the problem in `sails-postgresql` where passing `null` then gets converted to a NaN which breaks the query


weyert:
OK. I will see if I can get that going. Has been a while, haha!

CWyrtzen:
Bumped~

mikermcneil:
@weyert yeah, sorry about that- thanks for getting back :)  We've been working on the way we tackle issues and PRs in the project.  If you'd like to be a part of the "issue street team", let me know :)

Issue #2051, opened 2014-08-04
Don't overwrite app settings in middleware order


Kikobeats:
I'm trying and the core don't overwrite the app settings in middleware order

related: https://groups.google.com/forum/#!topic/sailsjs/MXGTPc7Jnic

PR #2083, opened 2014-08-08
[Feature Request]  My proposal of plugins system

Kikobeats:
Hello everyone,

This weeks I'm working trying to design a plugin system for Sails. At the beginning part of the [Yomguithereal](https://github.com/balderdashy/sails/pull/1522)'s concepts but early I detected other things that plugin system needed to support. (In my opinion, of course)

For this reason, I created a parallel proyect called [Sailor](https://github.com/sailorjs) which is my playground scenario to try out the plugin system where I try to improve it.

My idea is not to create a parallel proyect. The idea is that sails implements a plugin system and uses sailor for this.

I want to explain how the plugin system works and which modifications in the current sails version are necessary for this.

Hipster mode on, get ready for reading and GO!

## The idea

- We have an application and we can divide it in two parts: a unique core (called base), and one or more modules.
- The modules have logic domain that the core loads and they are a little piece of domain code.
- The responsability of each module is the module.
- The modules can be relationated between them.
- The modules can be combinated for create a custom backend.
- The modules follow [FIRST](http://addyosmani.com/first/) principles.

### Scaffolding support

Sails has a default configuration when you create a sails application. The sails application have a scaffolding that overwrite the start sails core settings.

The concept of the plugin system is the same, but with another depth level: sails has a configuration that lets you overwrite your base application, and you can overwrite or add more logical code with each module as well.

It could exist conflicts between base and module configuration. In this case, the base configuration has priority.

For this, I modified the core (`moduleloader` hook). I declared the modules in a new file called `config/plugins.js` that is an array of plugin namen that checks if it exists in `node_modules` as dependency. If it exist, I load it:

### Multilanguage support

I created a module called [sailor-translate](https://github.com/sailorjs/sailor-translate) that is based in the angular-translate solution. Basically it creates a JSON for each language and it is very easy to create and use (much more than locales files)

Why? Because for me is very important to create a multilang API from the beginning. You should work hard in your API development in order to make your frontend much easier to use.

You can check that this [pseudo-file](https://github.com/sailorjs/sailor-module-user/blob/master/translation/User.tl) compiles to [this one ](https://github.com/sailorjs/sailor-module-user/blob/master/config/translation.coffee). I use sailor-translate for [register the translation files](https://github.com/sailorjs/sailor-module-user/blob/master/config/bootstrap.coffee#L26) and [make it easier to access](https://github.com/sailorjs/sailor-module-user/blob/master/api/services/protocols/local.coffee#L31).

Another thing about this is that I had the core for creating route binding for each language (`blueprint` hook). I thought that it could be improved because in my solution I read the local files to know the languages and later I bind the routes for each language. I try binding a route like `/:lang/` but it doesn't work. Also, [I created a little middleware for fix language in the route](https://github.com/sailorjs/sailor-module-user/blob/master/api/policies/i18n.coffee) because some routes don't have a `/lang/` param.

### Tasking support

Tasks for each module are necessary because the modules are independent. For this reason I modified the core to load the module's tasks in a different thread compared to the base tasks thread. (`grunt` hook).

The only problem in this aspect is that you want to install a module as a dependency, before that you lift the server you need to install the local module grunt dependencies or the tasks that don't work (because the tasks are local in the module). In my launcher it doesn't work but is easy to fix.

## Scaffolding

The Scaffold of each module is similar to the base, but more extended:

```
--sails-module-example/
----api/
------api/adapters/
------api/blueprints/
------api/controllers/
------api/hooks/
------api/models/
------api/policies/
------api/responses/
------api/services/
----views/
----config/
------config/translation.js
----translation/
-----Gruntfile.js
```

In this aspect, I think that to overwrite blueprint is SO USEFUL to avoid rewriting code. Another think like adapters or hooks are not important for module design.

## One more thing

Another thing that I did not commented and that is a problem for my workflow is the waterline errors schema. It is ugly, and of course, it is not multilanguage. I propose a more simple schema based in [express-validator](https://github.com/ctavan/express-validator):

It is simple and clean. And the most important: [easy to setup](https://github.com/sailorjs/sailor-module-user/blob/master/api/services/protocols/local.coffee#L32) (and easy to make multilanguage). This is ready for frontend. Doesn't exist a waterline errors validation interface for validate fields out of the ORM.

## What To Improve

* Fixed grunt dependencies modules installation before lift server.
* Better way to binding multilang routes.
* Native mutlilang route support in the core and not use a middleware for this.
* Improve waterline errors and make multilang

## Demo

You can experiment with this features NOW.

1) Install [sailor](https://github.com/sailorjs/sailorjs) globally. Basically is a little CLI interface to run the sails core modification.

2) Check [sailor-module-user](https://github.com/sailorjs/sailor-module-user) that is a version of [sails-generate-auth](https://github.com/kasperisager/sails-generate-auth) rewrite like a module. (I have a little problem with login methods. Check the tests, try to fix and PR pls :P)

3) [sailor-app-playground](https://github.com/sailorjs/sailor-app-playground) is a little app that use this module. Experiment with this!

## Your turn: Feedback!

Please, now is your moment. Feedback this plugin system. For make a better Sails core.


loicsaintroch:
@mikermcneil @sgress454 Can you give us an update about this feature since you are both currently working on a subapp hook and a new way to load hooks?

dan-omniscience:
+1

thiagof:
Adding ideas:

I think a plugin should be able to create the same behavior of an app.
Idealy, it would have an `api` folder which triggers sails behavior
just like an app.

In this scenario not only extensions would be possible, but really
reusable projects and components.

Also would be trivial to refactor code into a plugin/component, thus
being easy to read and maintain by any sails app developer.

2015-01-04 19:46 GMT-02:00, dan-omniscience <notifications@github.com>:
> +1
>
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/balderdashy/sails/pull/2083#issuecomment-68650602


-- 
*Thiago Fernandes*
Analista de Processos
34 9176- 4055 | thiago@internetbudi.com.br

NathanRSmith:
Really nice pull request @Kikobeats. It would be great if this could be integrated soon. I'm needing exactly this functionality to continue using Sails.

loicsaintroch:
@Kikobeats Can you tell us the differences, the pros and the cons of your pull request instead of using the `subapps` hook? @sgress454 wrote this hook and it works perfectly.

albertosouza:
@loicsaintroch some time ago I created a hook for sails.js ( 0.10.x ) who has this feature and now I'm upgrading to sails.js 0.11.x, link: https://github.com/wejs/we-plugin

the idea is to work as plugins like Drupal, Joomla, Wordpress or Plone.

we-plugin is only a sails module loader with some "fix" in current version to work with sails 0.10.x how will be upgrated to sails 0.11.x and i have one small project working with it in https://novo.atencaobasica.org.br/

I guess you could consider the we-plugin as an option to work with plugins in the sails.js ....

Kikobeats:
@loicsaintroch I feel that are very different and for different purpose. 

I don't remember now how works @sgress454 hook system, but the idea behind this plugin system is, for each plugin, load a sails tree directory when your `lift` your application. 

Because when I'm working with sails I feel that I have to rewrite the some code in some projects. Wrong. Instead of I want to encapsulate the code, or common parts of code and re-use in different projects, and make easy adapt the code for each specific situation.

On the other hand, I added some features that support they system:

* A minimal core API for know plugins loaded, blind plugins that depend of other plugins.
* Because we are defining things in executing time we can do things very powerful as database schema conditional.

I collected all features in a [little book](http://kikobeats.gitbooks.io/sailorjs/content/es/) that is in spanish only at this moment. If exist people in the fork I can translate it quickly :-) (I should!)

And things that are not directly have relation with the plugin system, but I feel that I need in my workflow:

* When I put my code in production, my assets are ready, don't need to compile again in the cloud. This is very costly for the production machine and easy to resolve locally.
* Relative to the previous point, I can convert `grunt` dependencies into dev-dependencies and make the core build more lightweight (and you can use another workflow for assets, for example with `gulp`).
* A better (in my opinion) [multilanguage system](https://github.com/sailorjs/sailor-translate) based in javascript object instead of JSON. Is the same than angular translate.
* An ecosystem testing system for plugins. The CLI generate all that you need for the testing. You don't need to install the core but you can referenced it.
* Again, IMHO, if sails is a API-oriented framework is a must support tokens for authentication instead of cookies. [I have a little playground for this ideas](https://github.com/sailorjs/sailor-module-authentication)

and more, more things. I recommend check all [sailor repositories](https://github.com/sailorjs).

I don't think that sails team go to integrate this, at this moment is a PR with too many changes, but if exist the enough people interested, we can define a roadmap and prepare sailor for production.

What do you think?

sgress454:
@Kikobeats @loicsaintroch @albertosouza I know I'm weighing in on this a little late.  The fact is that there is no perfect solution for plugins in Sails, or even a [perfect consensus on what "plugins" are](http://www.thedotpost.com/2014/11/mike-mcneil-pulling-the-plug).  The [subapps](https://github.com/balderdashy/sails-hook-subapps) hook is intended more than anything as a demonstration of how powerful hooks can be in extending Sails without modifying the core.  

In coming releases, you'll see more and more core hooks separated into their own modules (the [sockets hook](https://github.com/balderdashy/sails-hook-sockets) is already a separate module in v0.11).  Most of the plugin ideas we hear for Sails could (and should) be implemented in v0.11.x as hooks without needing to load whole "slave" Sails apps as both [Sailor](https://github.com/sailorjs/sailorjs) and [subapps](https://github.com/balderdashy/sails-hook-subapps) do.  It's just rare to find a module idea that a) requires bringing the full force of Sails to bear and b) can be made generic enough to act as a plugin to other apps.  Building a generic admin panel for Sails apps comes to mind, or a CMS like [what @loicsaintroch is working on](http://extens.io/).

The Sails core team hasn't given Sailor the attention it probably deserved, because it was focused on modifying the core at a time when we were actively trying to make the core leaner and more modular.  It also seemed to rely on using the Sails module loader to load the plugin controllers, models, etc. directly into the host Sails app, which in my experience makes writing the plugins difficult because you have to avoid name collisions between the plugin and host (this may not still be how Sailor works, it's just what I recall from the first pass).  Now that Sailor is using a forked Sails, it's hard for us to actively support it, as much as we appreciate @Kikobeats and his [contributions](https://github.com/balderdashy/sails/issues/1598) to [Sails](https://github.com/balderdashy/sails/issues/2051).  Ideally the best ideas from Sailor could be merged back into Sails or (better) converted to hooks, making the framework stronger.  We're all after the the same thing, after all: making life easier and more fun for Node developers. :two_men_holding_hands::two_women_holding_hands::couple:

albertosouza:
@sgress454 thanks for this comment

And I am working to create a blogging and site generator with we.js and an admin like Wordpress but with SPA structure ... I think it may be related to the @loicsaintroch project. I will try to talk to him

mikermcneil:
Thanks for the great discussion everyone.

PR #2173, opened 2014-08-31
Add blueprintType argument and core support for ember

Globegitter:
It would be super nice to have out of the box support for JSONApi/ember-data.
I know you can manually overwrite blueprints, which works pretty good but it would be even better to have first class support for ember.

I have just started very basic by taking the ember blueprints from https://github.com/mphasize/sails-ember-blueprints (which are not a 100% complete, but a really good start), and adding a blueprintType setting to be recognized from the `config/bleuprints.js`.
I have split up the current blueprint actions into a new `default-actions` folder and added an `ember-actions` folder and depending on the setting either the default actions or ember actions get loaded in the hook.

Furthermore I started adding a `blueprintType` argument to `sails new` which currently doesn't so anything, but would then just set the right settings.

This PR is quite basic so far, since the questions poses if you would even want that functionality in core?
And then on how exactly this should work/if my approach is a good one? (Since I just started looking into the core for the first time).

If you that is something you would implement into the core I would then also create a PR for the `ROADMAP.md`. Just thought it would be better to first start the PR with some code ;)

So yeah let me know what you think.


sgress454:
[Sane Stack](http://sanestack.com) FTW!

PR #2206, opened 2014-09-10
Use 'www' folder if is available in production

Kikobeats:
I would like know the opinion about this.

Instead of define again a gunt task to copy the files in '.tmp/folder', use www/ folder with static assets compiled.

.cc @davidrivera

### DIFF
--- a/lib/hooks/http/index.js
+++ b/lib/hooks/http/index.js
@@ -2,10 +2,10 @@
  * Module dependencies.
  */
 
-var path = require('path');
-var _ = require('lodash');
+var path    = require('path');
+var _       = require('lodash');
 var express = require('express');
-
+var fs      = require ('fs');
 
 module.exports = function(sails) {
 
@@ -44,7 +44,7 @@ module.exports = function(sails) {
         //  • an absolute path
         //  • a relative path from the app root (sails.config.appPath)
         paths: {
-          public: '.tmp/public'
+          public: config.environment === 'production' && fs.existsSync('www') ? 'www' : '.tmp/public'
         },
 
 


Globegitter:
Yep, just running into a similar issue over at Sane Stack: https://github.com/artificialio/sane/issues/51

The grunt hook is disabled per default, because using docker, or nginx+pm2 (multi-process apps) it really is just causing issues, but I didn't think about the fact that the assets do not get lifted anymore on a simple `sails lift`.

This would certainly be an option, or how about having an option in the `.sailsrc`? Something like `assets: assets` or `assets: www`, etc? If it is empty it just defaults to `.tmp/public`. This makes it simple, new people don't get confused because nothing changes for them and others can customize it.

What are your thoughts?

mikermcneil:
@Globegitter agreed- so as you probably know, you can do this now by setting `sails.config.paths.public`- is the issue that it isn't being pulled from the sailsrc file in time to affect the static middleware? Or is it that the path isn't being resolved if it's relative?

ie. in your sailsrc file:
```json
{
  "paths": {
    "public": "./www"
  }
}
```

CWyrtzen:
Bumped~

randallmeeker:
What exactly does the "bumped" mean?

CWyrtzen:
@randallmeeker et. al.

It means we are doing what we can to move through the backlog of PRs and get caught up~ 

bump (verb)
past tense: bumped; past participle: bumped
to push (something) jerkily in a specified direction.

Hopefully we won't do it jerkily but I can't make any promises ;)

Kikobeats:
@CWyrtzen wat? and what's ~ means?

seriously, bumped doesn't mean nothing, and at the end you says *I can't make any promises*, so basically you are saying nothing of nothing...

Chocolate~

randallmeeker:
I'm still confused. Does it mean your doing something with this issue
specifically?

I know what it means for a community member to "bump". They are promoting
an issue or request.

CWyrtzen:
Twas a bit of a joke about not making promises...the note simply means we sorting it through, moving it along intentionally, taking a second and third and fourth look...

All I can say is we are doing what we can to catch up and clean up and shine up so when you see BUMP it means it has been hand-carried on your behalf.

Perhaps my attempt at humor is misplaced but I think a little levity and good will are essential in times like these.

And @randallmeeker yes, promote, absolutely: further the progress of (something, especially a cause, venture, or aim); support or actively encourage. Indeed~

Cheers~

mikermcneil:
Alright, closing in favor of the solution documented here: https://github.com/balderdashy/sails/pull/2543#issuecomment-69277536

PR #2274, opened 2014-09-30
Removes the socket hook's dependency upon the session hook.

lwansbrough:
So this removes the session hook requirements from the sockets hook. Helpful for those of us using JWTs.

### DIFF
--- a/lib/hooks/sockets/lib/authorization.js
+++ b/lib/hooks/sockets/lib/authorization.js
@@ -19,6 +19,8 @@ module.exports = function(sails) {
 
 	return function socketAttemptingToConnect (handshake, accept) {
 
+		if(!sails.config.hooks.session) return accept(null, true);
+
 		// If a cookie override was provided in the query string, use it.
 		// (e.g. ?cookie=sails.sid=a4g8dsgajsdgadsgasd)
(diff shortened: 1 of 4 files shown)


loicsaintroch:
Any update on this pull request guys?

mikermcneil:
@loicsaintroch @lwansbrough I think this is sorted in https://github.com/balderdashy/sails-hook-sockets/blob/master/lib/initialize.js#L50 - would you mind installing 0.11.0-rc5 and giving it a go?

lwansbrough:
Thanks Mike, I'll take a look a little later.

mikermcneil:
@lwansbrough alright I believe we're good now.  Added tests here: https://github.com/balderdashy/sails-hook-sockets/blob/master/test/without-session.test.js  Going to go ahead and close- thanks for your help :)

lwansbrough:
@mikermcneil Oops forgot to report back! Works great thank you.

fabdrol:
@mikermcneil Does this require 0.11?

tarlepp:
@fabdrol yep it requires 0.11.x sails, I just tested this with sails 0.11.0 and session hook disabled and it works like a charm.

PR #2285, opened 2014-10-02
Add a config variable to tell Sails if it should run the prod grunt task...

### LINKED ISSUE #2270: Remove invoking grunt:prod on production mode
Hi all. 
It's more a question to community.
May be we could remove invoking grunt when running sails in production mode and run grunt commands manually ?

Problem for me is this: 
I'm using sails.js v0.10.5 in production on live website.And using CI service. 
Also my application has lot of jade, js and css files for frontend.
So when I need to restart my app or make a small change into sources it it compiles all this resources every time even if I don't need it. 

For my installation I made a hook and disabled grunt for production. 
But might be it will be a good way to for this for others ?

What do you think ?

m-bodmer:
... when lifted in production mode. Fixes #2270 

In `config/local.js`, we could add the following snippet which would come from the generator:

```
/***************************************************************************
    * In production, Sails will run the 'prod' grunt task when the            *
    * server is lifted.                                                       *
    *                                                                         *
    * If you prefer to build your assets before lifting your server,          *
    * then set disableProductionBuild to true.                                *
    ***************************************************************************/
   // disableProductionBuild: true
```

Please let me know if there are any comments/concerns

### DIFF
--- a/lib/hooks/grunt/index.js
+++ b/lib/hooks/grunt/index.js
@@ -21,9 +21,15 @@ module.exports = function (sails) {
 
       sails.log.verbose('Loading app Gruntfile...');
 
-      // Start task depending on environment
-      if(sails.config.environment === 'production'){
+      /* Start task depending on environment
+       * Run the prod task by default when Sails is running in a 'production' environment
+       * If disableProductionBuild is set to true in config/local.js, then do not run this task
+       */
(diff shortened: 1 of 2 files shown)


particlebanana:
You can turn the grunt hook off in `sailsrc` as seen [here](https://github.com/balderdashy/sails/pull/2543#issuecomment-69277536). This would also still run the `default` task.

PR #2304, opened 2014-10-10
custom view extensions

mrfishie:
Adds the 'sails.config.views.extension' config property that allows for a custom view file extension easily


cage1618:
@sgress454  This feature had  removed in v1.0?

PR #2355, opened 2014-11-03
Check if association.alias is set before checking the length

erkstruwe:
same as in association.type == 'model'


sgress454:
@erkstruwe I think what @mikermcneil is saying is that you probably want:

```
else if (association.type == 'collection' && association.alias && previous[association.alias] && previous[association.alias].length) {
```

rather than `[assocation.alias]`, since that is always truthy and therefore doesn't contribute anything to the conditional.  Also: we are always in favor of safer dereferencing, but is there a test case where something fails without this?

CWyrtzen:
@erkstruwe is there any progress for this?

erkstruwe:
Sorry guys. Had a lot to do and was on vacation in the meantime... I looked everything up in my code: `previous[association.alias] && previous[association.alias].length` was the correct check I implemented. So even my PR is faulty...

In addition, I really can't remember the use case for this. But I'm sure it was some weird usage during my early days of Sails development... Due to all of this, I'm just closing this issue.

Issue #1959, opened 2014-07-14
Sails 0.10.10-rc8 / mongo session: receives incoming connection before connect-mongo is connected


fabdrol:
I'm getting this error - but only sometimes: it occurs when I have a page open in the browser (which uses sockets) or if I navigate to that page too early. It seems to be caused by connect-mongo receiving an incoming session connection before connect-mongo has managed to connect to the database. See [this question on SO](http://stackoverflow.com/questions/22698661/mongodb-error-setting-ttl-index-on-collection-sessions)

```
node_modules/connect-mongo/lib/connect-mongo.js:161
                throw new Error('Error setting TTL index on collection : ' + s
```

mikermcneil:
@fabdrol ok looks like we just need a quick rebase, let's continue discussion in the PR: https://github.com/balderdashy/sails/pull/2407  

thanks again!

PR #2407, opened 2014-11-17
CB for connect-store, so mongo-store will only cb upon conn complete

bakavic:
Following the example provided here: http://stackoverflow.com/a/22698992/3440368

For this issue : https://github.com/balderdashy/sails/issues/1959

I'm not sure if this is the right fix - the tests pass, but would appreciate if a developer could guide me through :) (Namely, I'm not sure how to write a test for this.)


mikermcneil:
@bakavic Thanks Victor-- I did a bit a work on this recently while massaging the virtual request interpreter for v0.11, so going to to have to cherry pick this one.  I think the mongo store issue is still a problem with the improved code though...  If you don't have time to rebase, np (myself or someone else will get to it soon), but would you mind having a run at it w/ the latest code to double-check that the issue is still the same?

/cc @fabdrol @luislobo

luislobo:
I moved the session to Redis, and I didn't have any more issues.

bakavic:
@mikermcneil I didn't experience the problem personally, so I can't reproduce it.

I have done the rebase (not sure why github brought in all the other commits - am I rebasing wrongly?), but in any case, this fix doesn't work with `connect-mongo` > 0.4.2 -  [MongoStore no longer accepts a callback](https://github.com/kcbanner/connect-mongo/commit/ca40b7848c1791362da4e4e9890dae6dcb079b62#diff-3a4675a32773eb2f52151d91494b93f1L63). 

That commit claims to fix the [issue](https://github.com/kcbanner/connect-mongo/issues/70), so this PR can be discarded, and testing should be done to see if the issue is finally solved.

CWyrtzen:
Bump~

sgress454:
It's hard at this point to tell what the actual changes from this PR were.   Closing this for now; if someone wants to write tests to try and reproduce the original issue, that'd be grand.

PR #2435, opened 2014-11-24
Fix a bug when using 'where' with a datetime in Blueprints

Dallas62:
Hi!
Values are not parse well when you use blueprint,
if you run the `npm test` of an adapter, you will have greaterThan and lessThan tests on Dates. Those tests generate data and when you run the adapted test for blueprints:
`http://localhost:1337/api/user2/?where=%7B%22type%22%3A%22lessThanOrEqual%20dates%20test%22%2C%22dob%22%3A%7B%22%3C%3D%22%3A%222013-11-01T23%3A00%3A00.000Z%22%7D%7D`
from 
```
        it('should return records with lessThanOrEqual key when searching dates', function(done) {
          Queryable.User.find({ type: testName, dob: { lessThanOrEqual: new Date(2013, 10, 2) }}).sort('first_name').exec(function(err, users) {
            assert(!err);
            assert(Array.isArray(users));
            assert(users.length === 2);
            assert(users[1].first_name === 'lessThanOrEqual_dates_user1');
            done();
          });
        });

        it('should return records with symbolic usage <= usage when searching dates', function(done) {
          Queryable.User.find({ type: testName, dob: { '<=': new Date(2013, 10, 2) }}).sort('first_name').exec(function(err, users) {
            assert(!err);
            assert(Array.isArray(users));
            assert(users.length === 2);
            assert(users[1].first_name === 'lessThanOrEqual_dates_user1');
            done();
          });
        });
```
The result is an Array of ONE model ( TWO expected ). (with sails-mysql and maybe other)

This fix is not THE answer to fix this, but it's work.

And replace tryParse by sails-util.tolerantParse.

### DIFF
--- a/lib/hooks/blueprints/actionUtil.js
+++ b/lib/hooks/blueprints/actionUtil.js
@@ -2,9 +2,10 @@
  * Module dependencies
  */
 
-var _ = require('lodash'),
-  mergeDefaults = require('merge-defaults'),
-  util = require('util');
+var _             = require('lodash'),
+    mergeDefaults = require('merge-defaults'),
+    util          = require('util'),
+    sailsUtil     = require('sails-util');
 
 
 // Parameter used for jsonp callback is constant, as far as
@@ -179,7 +180,7 @@ module.exports = {
 
     // If `where` parameter is a string, try to interpret it as JSON
     if (_.isString(where)) {
-      where = tryToParseJSON(where);
+      where = sailsUtil.tolerantParse( where, dataReviver );
     }
 
     // If `where` has not been specified, but other unbound parameter variables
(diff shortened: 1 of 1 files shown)


Dallas62:
@mikermcneil
Take a look at https://github.com/lapwinglabs/superjson that can solve this issue.
(Or similar)

CWyrtzen:
Bump~

sgress454:
Thanks for the report.  This may have been fixed at the Waterline or adapter level; I can't reproduce after testing with sails-postgresql, sails-mysql and sails-disk.

Dallas62:
Hi @sgress454,
Will check soon, but was a Blueprints bug, not a adapter bug.

mikermcneil:
@Dallas62 ah @sgress454 was just clarifying that he tested all three adapters just in case.

mikermcneil:
see https://github.com/balderdashy/sails/pull/2445

PR #2445, opened 2014-11-26
Exception for invalid JSON in req.options.where

gadelkareem:


mikermcneil:
@gadelkareem  apologies for the delay.  This was resolved by https://github.com/balderdashy/sails/pull/2435 afaik, but please resubmit if your use case wasn't taken care of.

Issue #2449, opened 2014-11-27
Blueprints doesn't support multiple sort


Dallas62:
Hi!

Actualiy, blueprints doesn't parse `sort` argument as json or by splitting string,
so we can't do something like:
`/model/?sort=foo ASC, bar DESC`
`/model/?sort=[ 'foo ASC', 'bar DESC']`
`/model/?sort={ 'foo': 'ASC', 'bar': 'DESC'}`
`/model/?sort={ 'foo': 1, 'bar': 0}`

His there any purpose to do it well?
Maybe a PR soon.

sgress454:
Good idea @Dallas62.  Added support for JSON sort options in https://github.com/balderdashy/sails/commit/636284e7ebe0382c9fc5e335333de22a58135b99

Note it must be proper JSON (with double-quotes) for it to work.

sgress454:
Somehow missed that you had a PR to do this exact thing.  Sorry about that, wasn't intentional.

PR #2450, opened 2014-11-27
Add multiple sort support on blueprints

Dallas62:
Hi!

This PR add multiple sort support on blueprints

Basic usage still working:  `?sort=foo ASC`
but add JSON support like `?sort={foo: 'ASC', bar: 'DESC'}` (basically same parameters than Waterline)

As referred https://github.com/balderdashy/sails/issues/2449

### DIFF
--- a/lib/hooks/blueprints/actionUtil.js
+++ b/lib/hooks/blueprints/actionUtil.js
@@ -274,7 +274,18 @@ module.exports = {
    * @param  {Request} req
    */
   parseSort: function (req) {
-    return req.param('sort') || req.options.sort || undefined;
+    var sort = req.param('sort');
+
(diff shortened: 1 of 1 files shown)


kevinburkeshyp:
Should prolly add a test for this

tjwebb:
@Dallas62 can you include a test that shows how this works? thanks

Dallas62:
@tjwebb We are leaving Sails from our projects, I don't have enough time for that.
Data for integration tests doesn't help to quickly include test. There is only properties with the `userId` so I can not use them for more than one sorting arguments.

I tested it with `waterline-adapter-tests` and a custom Waterline adapter based on Blueprint and I will probably publish it soon.

Cordially.

PR #2463, opened 2014-12-04
Layout support for Ractive templates

sleewoo:
Basically rendering a layout with given string as `body` partial.

If no layout defined in `config/views.js`, `{{> body}}` used for layout.
Also ad hoc layouts allowed: `res.view(layout: 'some-layout')`

Layout is read from fs and cache used on production.


loicsaintroch:
Thank you for your pull request @sleewoo. Can you also update `configure.js` with better logs and `layoutshim.js` only if necessary please?

sleewoo:
@loicsaintroch, updated `configure.js`
There are no loadtime setups for ractive layouts for now, so no need to update `layoutshim.js`.

loicsaintroch:
Thank you @sleewoo

sleewoo:
happy to contribute. Sails is amazing. inspired me a lot.

PR #2465, opened 2014-12-05
pull apart actionUtil.populateEach to support reuse outside requests

askmike:
Turns actionUtil.populateEach into 3 different functions that craves the way towards more reusable functions.

Our use case:

In certain specific scenarios [we want to publish actions](https://stackoverflow.com/questions/27254587/sails-publishupdate-system-does-not-propagate-to-associations) (publishUpdate et al.) to associated models (eg: User has Pets, when a pet changes, publish User update). We want to auto populate the associated models but where unable to use populateEach due its dependence on a `req` object. 

EDIT: next step would be exposing those helper function to userland somehow.

### DIFF
--- a/lib/hooks/blueprints/actionUtil.js
+++ b/lib/hooks/blueprints/actionUtil.js
@@ -18,18 +18,19 @@ var JSONP_CALLBACK_PARAM = 'callback';
  *
  * @type {Object}
  */
-module.exports = {
+var actionUtil = {
 
   /**
-   * Given a Waterline query, populate the appropriate/specified
-   * association attributes and return it so it can be chained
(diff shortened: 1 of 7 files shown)


sgress454:
It would be nice to have some tests for these.  This in particular will fail with globals turned off:
https://github.com/balderdashy/sails/pull/2465/files#diff-8203dfaef1a1458ab9cdba0187da0550R106

I patched that (and other issues pertaining to using globals in the core) here: https://github.com/balderdashy/sails/blob/189937df23883523823d9c02929b1b358410ca18/lib/hooks/blueprints/actionUtil.js#L105-L106

Issue #2473, opened 2014-12-06
[Question] Better way to implement blueprint prefix configuration.


konstantinzolotarev:
Hello. 
I want to enchase blueprint configs a little and add an ability to set prefix for rest actions. 

For now we have `prefix` option into [blueprint config](http://sailsjs.org/#/documentation/reference/sails.config/sails.config.blueprints.html) but it set prefix for all blueprint actions including controller.action, shortcuts and rest.

But In my application I need to be able to use controller.action without any prefix but RESTful routes should be prefixed with `/api/` prefix. 

I could add a new option named `restPrefix` with same logic as `prefix` but only for RESTful routes.  But if someone will use both of this options. for example: 

```javascript
module.exports.blueprints = {
  prefix: '/prefix',
  restPrefix: '/api'
}
```

How should application route this ? 
Should all RESTful routes be accessible via `/api/...` or it should be accessible  via `/prefix/api/...` ?

kyle-schmidt:
@konstantinzolotarev this isn't working properly on Windows as path.normalize converts / into \ on Windows, making the routes not work correctly.  Works great otherwise.

konstantinzolotarev:
@kyle-schmidt Fix is there. waiting for merge.
https://github.com/balderdashy/sails/pull/2666

Issue #2031, opened 2014-07-30
Suggestion: support for multiple blueprint prefixes


mnaughto:
I was wondering if there is any interest in adding different blueprint prefixes (e.g. one for actions, one for RESTful endpoints).

I tweaked /lib/hooks/controllers/index.js to allow an "actionPrefix" in addition to a regular "prefix". This way, I can have RESTful actions under an "api" prefix for all of my models, but I can still get the automatic route binding for controllers that deliver HTML to the client with an empty prefix.

Since I'm on 0.9, this might be completely irrelevant in the new release...

PR #2501, opened 2014-12-21
RESTful actions prefix configuration in blueprints

konstantinzolotarev:
Fix for #2031 and #2473


sgress454:
Thanks @konstantinzolotarev, looks good!

PR #2502, opened 2014-12-22
Fix integration tests socket config for sails v0.11+

konstantinzolotarev:
debug: Deprecation warning: `sails.config.sockets.authorization` is now `sails.config.sockets.beforeConnect` (setting it for you this time)


konstantinzolotarev:
Not actual any more.

PR #2509, opened 2014-12-24
blueprint destroy action returns 204 status code

ligerdave:
Http status code 204 should be returned as one record getting deleted successfully

### DIFF
--- a/lib/hooks/blueprints/actions/destroy.js
+++ b/lib/hooks/blueprints/actions/destroy.js
@@ -40,7 +40,7 @@ module.exports = function destroyOneRecord (req, res) {
         }
       }
 
-      return res.ok(record);
+      return res.destroyed(record);
     });
   });
 };
--- a/lib/hooks/responses/defaults/destroyed.js
+++ b/lib/hooks/responses/defaults/destroyed.js
@@ -1,0 +1,48 @@
+/**
+ * 201 (DESTROYED) Response
+ *
+ * Usage:
+ * return res.destroyed();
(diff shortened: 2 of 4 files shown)


mikermcneil:
@ligerdave I think it has to do w/ a new default response being added- some of our tests could use better err msgs.  Seems like travis has sort of gotten "stuck" on this PR as well for some reason...

mikermcneil:
@ligerdave going to go ahead close this for now (btw, let me know if you're interested in helping to maintain the standalone blueprints hook as we work towards continued modularization of core for v0.12)

sgress454:
The tests were failing b/c this was submitted during a period where *all* tests were failing, due to the http server not shutting down in `sails.lower`.  That's been fixed, but in general I don't think we should be sending back a 204 code from the blueprint delete, because we're sending data back in the response indicating which record was deleted; 204 is supposed to indicate a successful response with no content, so that the client can be assured that everything worked even though they're getting no data.

mikermcneil:
@sgress454 @ligerdave also the 204 response would make a lot of existing tests and integrations with existing Sails apps fail, since they're checking for a 200 (which granted, should really be doing the standard `statusCode>=200 && statusCode<300`- but realistically we can't break existing apps without a _really_ good reason)

@ligerdave I really appreciate your time though; this is the perfect sort of thing that could be accomplished in a hook override (and potentially as continued extensions on top of the blueprint hook). In order to do that, we've got to pull these things out of core so that we're guaranteed to not have any entanglement

PR #2520, opened 2014-12-27
Fixed bug with sails.request params

Josebaseba:
I found that when I try to call in test to an url with sails.request, the params were ignored:

```javascript
   sails.request({
      url: '/id-required',
      method: 'post',
      params: {
        id: '1'
      }
   }, function(err, res, body){
      if(err) return done(err);
      body.status.should.be.equal(200);
      return done();
  });
```

The req.params.all() === {} in the controller. 

I fix it with just renaming the params var, and that's it!

### DIFF
--- a/lib/app/request.js
+++ b/lib/app/request.js
@@ -56,7 +56,7 @@ module.exports = function request( /* address, body, cb */ ) {
     url = arguments[0].url;
     method = arguments[0].method;
     headers = arguments[0].headers || {};
-    params = arguments[0].params || arguments[0].data || {};
+    body = arguments[0].params || arguments[0].data || {};
   }
   // console.log('called sails.request() ');
   // console.log('headers: ',headers);


mikermcneil:
@Josebaseba thanks!  Before I dig in- are you still seeing this on the latest npm-published beta version? (i.e. `npm install sails@beta`)

Josebaseba:
Yes sir, I've the latest version: 0.11.0-rc9, and the bug is there.

mikermcneil:
@Josebaseba thanks :+1:

eperedo:
Hi,

This method only works if we use **params.all()** in the controller?

```
   sails.request({
      url: '/my-method',
      method: 'post',
      params: {
        id: '1'
      }
   }, function(err, res, body){
      if(err) return done(err);
      body.status.should.be.equal(200);
      return done();
  });
```

If I try to get the parameter name using **req.params.id** it always get undefined. (v0.11)

Josebaseba:
@eperedo try this way: 
```javascript
  req.param('id');
```
check: http://sailsjs.org/#!/documentation/reference/req/req.param.html
req.params.id will work just by url params ->http://sailsjs.org/#!/documentation/reference/req/req.params.html

eperedo:
You're right **request.param()** works, thank you @Josebaseba!

PR #2531, opened 2015-01-05
More correct root path for ect template engine.

seth2810:
Now user can override views path inside config.


loicsaintroch:
Thank you @seth2810

PR #2533, opened 2015-01-05
req param default value

Enkows:
same with this https://github.com/balderdashy/sails/pull/2511


mikermcneil:
@Enkows thanks so much for the rebase :+1:

PR #2543, opened 2015-01-08
Assets folder is now customizable through the sailsrc file.

Globegitter:
That would be the most simple implementation of what I mentioned in https://github.com/balderdashy/sails/pull/2206

This would not allow folder names such as `false`.

Any thoughts?

### DIFF
--- a/lib/hooks/http/index.js
+++ b/lib/hooks/http/index.js
@@ -6,6 +6,7 @@ var path = require('path');
 var _ = require('lodash');
 var express = require('express');
 var methodOverride = require('method-override');
+var rconf = require('../../app/configuration/rc');
 
 
 module.exports = function(sails) {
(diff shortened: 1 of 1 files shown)


loicsaintroch:
Hi @Globegitter. I like the idea because it could be useful in certain cases. However it's kind of tricky to do that using `rc` and the `.sailsrc` file. For example, I never use the `.sailsrc` because I don't want/need it. I'm probably not the only one.

This customization is good but could be better as a basic configuration without using the `.sailsrc` file.

Globegitter:
Hi @loicsaintroch so you would prefer having a setting in one of the `/config` files? @mikermcneil @sgress454 any thoughts?

This would really help us keeping the grunt hook disabled but still serving the ember app via `sails lift`.

sgress454:
I might be misunderstanding your intent, but you can already change the assets folder location using regular configuration files or .sailsrc, by setting `paths.public`...

Globegitter:
Aahh that is almost what I thought and might be all I need. I'll try it out and get back asap.

Globegitter:
Oh yeah this already works with:
```js
{ 
  "hooks": { 
    "grunt": false 
  },
  "paths": {
    "public": "assets"
  }
}
```

Just does not seem to be documented. Will try to do asap.

PR #2572, opened 2015-01-20
Update async dependency

andyhu:
The version included in sails is too old, it lacks some functionality like `async.retry` etc.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
   },
   "dependencies": {
     "anchor": "~0.10.0",
-    "async": "~0.2.9",
+    "async": "~0.9.2",
     "captains-log": "~0.11.8",
     "colors": "~0.6.2",
     "commander": "~2.1.0",
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
   },
   "dependencies": {
     "anchor": "~0.10.0",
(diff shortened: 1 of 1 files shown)


boogerlad:
This build has failed because https://github.com/caolan/async/issues/695

andyhu:
@boogerlad Thanks for the info. I changed the dependency to v0.9.0 and it passed the test.
async and lodash are exposed as a global variable so I think it's necessary to provide developers with up to date library, otherwise sometimes it's a bit annoying that something will break just because the library is too old and not consistent with the current documentation

mikermcneil:
@boogerlad thanks for the catch

> it's a bit confusing that something will break just because the library is too old and not consistent with the current documentation

@andyhu agreed!  The `async` and `_` globals are really just there for convenience but it is absolutely pretty weird if they are tangibly different.  Thanks for sending this :+1:

mikermcneil:
@andyhu would you mind taking responsibility for keeping an eye on the async dep for the project in the future?  (also feel free to hit me up at [@mikermcneil](http://twitter.com/mikermcneil) as well if I miss your pr next time)

andyhu:
@mikermcneil Sure, will do that with pleasure :)

Issue #2573, opened 2015-01-20
PUT is not support  in the association between two records 


shanelau:
As the document says,

Remove from Collection
Removes an association between two records.

```
DELETE /:model/:record/:association/:record_to_remove
```
GET、POST、DELETE are support by sails，why is there no PUT?

PUT  /:model/:record/:association/:record_to_remove ，the result is 404

sgress454:
If we have a `PUT` method on an association route like `/user/1/pets/2` where we already have `POST`, the implication is that you can modify the specified record in place.  But you can already do that with a `PUT` request to `/pets/2`; there's no need for a new route.

I think there might be some confusion because we offer the `GET` route which also appears to be redundant; why do `GET /user/1/pets/2` when you can just do `GET /pets/2`?  The answer is: you shouldn't\*.  In fact, you get less information that way because the former route won't automatically populate associations.  What we're really providing is `GET /user/1/pets`, which provides a list of *all* pets belonging to the owner without the owner information itself; the longer route just sort of tagged along for the ride.

<hr/>

<sub>*unless you don't have unique IDs for pets...in which case...you should have unique IDs for pets.</sub>

shanelau:
This confused happen in using angular-data with sails.

 In angular-data , update some sub data，

```
DS.update('comment', 1, { content: 'stuff' }); // PUT /post/2/comment/1
```

Sails dose not support PUT method, so I have to use 
`$http.put('/comment/1')`

PR #2574, opened 2015-01-20
Adding the route for PUT in association blueprints

Dallas62:
Adding `PUT /:model/:record/:association/:id` to add an association

fix this https://github.com/balderdashy/sails/issues/2573


sgress454:
See comment [here](https://github.com/balderdashy/sails/issues/2573#issuecomment-71166572).

PR #2599, opened 2015-01-30
Add IcedCoffeeScript support, add lazy loading for CoffeeScript and Iced...

internalfx:
...CoffeeScript

Here is a possibility for adding lazy loading for CoffeeScript and IcedCoffeeScript.

Not super happy about using synchronous APIs, but it looked like I would have to tear the whole thing apart to do it Async.

Would love feedback, thanks.


loicsaintroch:
I like it. I am down with that only if @mikermcneil and @sgress454 agree too.

sgress454:
@internalfx Definitely appreciate the work you put in here.  Merging this PR would be portentous of a future where `moduleloader` was 100,000 lines of code supporting every possible variation of Javascript.  What I would really love to see instead is a PR that rips all of the `|coffee|litcoffee|iced|liticed` lines out, along with the overhead of enabling CoffeeScript and its variants, and makes a modular system where you just `npm install sails-hook-icedcoffee` if you that's what you want to write Sails apps in.  It's not a trivial thing by any means, but I hope you can see how the course `moduleloader` is currently on is not sustainable.

internalfx:
I agree that its not sustainable forever, but it's probably sustainable for a couple years. I'm not too worried about there being 5 new JavaScript variants within a few years, but who knows?

But long term, you are absolutely right. It's just gonna bloat into a bigger and bigger mess.

`npm install sails-hook-icedcoffee` would be really sweet. So clean, keeps core small. I like it.

I'm not entirely sure where to start on that. I'm gonna need guidance from the core team. I'm willing to look into doing that though. The Sails project has helped me, I'm happy to give back.

I do however think there is some value here in the interim though. Not even just for `IcedCoffee`. This allows the logger to log an `error`, and not just `verbose`. It also will not error at all unless they try to use `Coffee` or `IcedCoffee`.

Thoughts?

Oh and `5` points to @sgress454 for use of the word "portentous" :smile:

mikermcneil:
@internalfx would you mind implementing this as a hook config?  i.e. replace the regexp def string with `sails.config.moduleloader.filter` here https://github.com/balderdashy/sails/blob/master/lib/hooks/moduleloader/index.js#L206

That way we can nip this one in the bud once and for all

internalfx:
@mikermcneil Happy to do that, but I'm a little lost....

`sails.config.moduleloader.filter` doesn't appear to be defined.
`sails.config.moduleloader` isn't even defined.

where do you want it? I haven't looked under the hood a whole lot.

Also, the regex isn't the same everywhere, I'm assuming you want a global config? To be DRY'er?

Compare these two lines...
https://github.com/balderdashy/sails/blob/master/lib/hooks/moduleloader/index.js#L206
https://github.com/balderdashy/sails/blob/master/lib/hooks/moduleloader/index.js#L243

Some have `|json|` in there....

How would you like me to proceed?

internalfx:
OK @mikermcneil . I had time to rethink my situation....

I was thinking about my response, and realized I should have put more effort into this. Sorry about that. Moving on...

So I "looked under the hood", and it seemed like the way you guys build `sails.config` is actually really simple. Pretty much wherever and whenever it suits you. Cool, I can do that.

> regex isn't the same everywhere

Right, so fix it. (Was the response I imagined you saying in my head)
So I implemented an array of extensions to pull in, and I build those into regexes where necessary.

OK, how's this look now?

internalfx:
@loicsaintroch @sgress454 @mikermcneil 

...Bumpity Bump....busy week?

lib/hooks/moduleloader/index.js
```diff
@@ -93,6 +80,56 @@ module.exports = function(sails) {
...
+      }
+      if (hasCoffee) {
+        try {
+           require('coffee-script/register');
+        } catch(e0){
+          try {
+            require(path.join(localConfig.appPath, 'node_modules/coffee-script/register'));
+          }
+          catch (e1) {
+            sails.log.error('Please run `npm install coffee-script` to use CoffeScript!');
```
kevinburkeshyp: s/coffescript/coffeescript

internalfx:
progress?

thoughts?

anything?

internalfx:
Ok, I have had my rant on #2662.

Shall I update this to add livescript support? so that a merge will please everyone.

Pleasing everyone was my original intent with this.

lib/hooks/moduleloader/index.js
```diff
@@ -109,6 +80,71 @@ module.exports = function(sails) {
           layout: path.resolve(config.appPath, 'views/layout.ejs'),
         }
       };
+
+      var hasCoffee = false, hasIced = false, hasLive = false;
+
```
internalfx: First we find out what languages are being used y walking the source files, and flagging what we find

lib/hooks/moduleloader/index.js
```diff
@@ -109,6 +80,71 @@ module.exports = function(sails) {
...
+          },
+          errors: function (root, nodeStatsArray, next) {
+            next();
+          }
+        }
+      };
+
+      walk.walkSync(localConfig.appPath+'/api', walkOptions);
+      walk.walkSync(localConfig.appPath+'/config', walkOptions);
+
```
internalfx: Then we only attempt to load the libraries if we found anything....

tjwebb:
@internalfx send me your email address when you get a chance re #2662 : tjwebb@balderdash.co

lib/hooks/moduleloader/index.js
```diff
@@ -117,6 +153,11 @@ module.exports = function(sails) {
       // Expose self as `sails.modules` (for backwards compatibility)
       sails.modules = sails.hooks.moduleloader;
```
internalfx: I used an array to build the regexes. So if we need to add another language later, we dont have to go edit EVERY regex in the file...

internalfx:
done, I added some line notes.

This *should* be ready for review

tjwebb:
Hey I just went through this PR, everything looks good.

@internalfx would you mind adding some documentation that explains the additional features that this provides, and some info on how to use it (or, if it's magic, then say so :) ).

@mikermcneil @sgress454 what are your thoughts? Things are always nicer as hooks, but I submit to you that existence is the more important feature in this case. Lots of stuff needs to be factored out into hooks anyway, and I think that will be a project in itself.

internalfx:
@tjwebb I'm working on cleaning this up some more....

tjwebb:
Ok thanks @internalfx. Ping me when it's ready!

internalfx:
Dont merge this yet...

internalfx:
oh wow...you reply fast

internalfx:
@tjwebb This is ready.

mikermcneil:
@internalfx Re: synchronous vs asynchronous loading-- I'm right there with you.  This is why we pulled out `sails-build-dictionary` back in 0.9-- to make the loading _potentially_ asynchronous.  But yeah, the standard answer from isaacs is that `require` is designed for synchronous use, and that it's ok because it only happens once.  It's really only a minor developer UX problem, but one of these days, I'd like to help on that front in node/npm, but for now we'll have to live with it.

Anyways, thanks!

internalfx:
@mikermcneil Yay! My projects can finally all run master again.

I wonder why require doesn't just take an optional callback? Such a system would keep all backwards compatibility.

On another topic, what are your thoughts on the idea of core hooks being override-able? (is that a word?)

Not sure how feasible that is, but then I could experiment with new core hooks. The `moduleloader` hook was so tied in that I couldn't make a nice clean way to add support for more languages. I would like to experiment with a custom `moduleloader`. (hope that doesn't make you cringe)

mikermcneil:
For an update on this, see https://github.com/balderdashy/sails/pull/3142#issuecomment-247690755  (let's funnel future discussion there as well so it's all in one place)

PR #2605, opened 2015-02-02
Added ability to pass an array of paths to sails.config.views.partials for handlebars view configurations

imartingraham:
The [express-handlebars](https://github.com/ericf/express-handlebars#partialsdirviewspartials) package allows for multiple partial directories.  Instead of passing a single string you can pass an array of strings.  This commit changes the ```layoutshim.js``` file which runs the handlebars view configuration. 

This functionality is beneficial in larger projects where grouping partials by controller makes more sense than putting all partials into a single folder. Here is an example use case:

```
── views/
   |── category
   |─── category_page.handlebars
   |─── partials
      |─── list_item.handlebars
      |─── left_nav.handlebars
   |── product
   |─── product_page.handlebars
   |─── partials
      |─── price_box.handlebars
      |─── description.handlebars

```

### DIFF
--- a/lib/hooks/views/layoutshim.js
+++ b/lib/hooks/views/layoutshim.js
@@ -2,7 +2,8 @@
  * Module dependencies
  */
 
-var path = require('path');
+var path = require('path'),
+		_ = require('lodash');
 
 
 /**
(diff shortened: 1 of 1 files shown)


loicsaintroch:
All right I see what you mean. I think this pull request also needs to update the `moduleloader` hook if we want to be specific and also the `sails-generate-backend` repository. Am I right?

imartingraham:
@loicsaintroch I'm not quite sure I understand the need for updating the ```moduleloader``` and the ```sails-generate-backend``` repo.  I'm fairly new to using sails so there is a good chance I'm missing how those pieces fit together.  I'm more than happy to make the changes but might need a little guidance on where I should go digging.

The change to the ```layoutshim.js``` file adds the ability to pass an array for handlebar templates.  You are still able to set a single string in the config.  This should allow the generators like ```sails-generate-backend``` to still work as they did previous to this change.

mikermcneil:
Thanks @imartingraham!  I'm not using handlebars in a project right now, so I don't have a solid existing code base to use as a test here.  @loicsaintroch, would you mind trying this patch out with one of your existing projects as a sanity check?

CWyrtzen:
@loicsaintroch were you able to take a look at this?

mikermcneil:
Hi @imartingraham!  I believe this should be relatively straightforward to take care of by customizing the `views` hook and including it as an override (either in `api/hooks/` or in your own package).  See https://github.com/balderdashy/sails-hook-orm for an example.

mikermcneil:
@imartingraham btw if you have any trouble with that, please don't hesitate to open a new issue-- happy to help with this if I can.  It's really important to me that we improve the process of doing these sorts of overrides: it's the best way for the feature set of Sails to scale beyond the use cases exposed out of the box by core.

PR #2606, opened 2015-02-02
remove grunt

boogerlad:
grunt should be a separate hook. I personally use gulp now. I can provide a gulpfile that is feature equivalent to the current gruntfile if needed. The dev task takes about half the time compared to grunt, while the prod task is roughly the same.


loicsaintroch:
I like the idea since a lot of developers are asking for it but the problem is I am quiet not sure we are ready to do such a break. In fact, removing the `grunt` hook will break every front-end workflow and every existing applications running on top of Sails. What do you think @mikermcneil @sgress454?

boogerlad:
Perhaps upon lifting, there should be a deprecation warning if a gruntfile is detected and say `run "npm install sails-grunt" if you want to re-enable the old build process"

boogerlad:
the hook now exists: https://www.npmjs.com/package/sails-hook-grunt
Caveats:
`sails www` is still gone
`test/integration/www.test.js` should be a test in the new hook?
grunt and associated plugins are not installed automatically to the project's node_modules folder.

boogerlad:
woops please ignore these commits. I've updated the master branch for my own personal use and it's a combination of my other pull requests.

mkeremguc:
+1 for gulp

sgress454:
@boogerlad a few questions on this:

1. Is this PR updated so that it only includes commits relevant to removing the Grunt hook?
2. What is your proposed alternative to `sails www`?   People do use that command.
3. How does sails-hook-grunt work with all of the Grunt plugins removed?  
4. Can you update the package.json of sails-hook-grunt so that it points to the correct repo?

I'm still onboard with moving this out of core, just need to make sure it doesn't break people's existing workflows.  Thanks!

mikermcneil:
@sgress454 @boogerlad OK- closing this for now.  It you're interested in helping extrapolate the grunt hook and update the relevant generators, please send a PR.  Thanks!

PR #2624, opened 2015-02-07
higher resolution request timer

boogerlad:


boogerlad:
Hmm, I seemed to have encountered a bug. When I do a socket request, it's getting interpreted as an http request!? In verbose mode, I see `Interpreting socket.io message as virtual request to "get [the url]"...
`

boogerlad:
Sorry if I'm being a pain in the ass, but can anyone chime in why socket requests are getting interpreted as http? I know that this `isSocket` worked in 0.10. @mikermcneil any ideas? I think you would know the most since you did most of the socket upgrading and http hook refactoring.

sgress454:
Part of the whole philosophy of Sails is that you can use the same backend code to handle both socket and HTTP requests.  As long as your socket request conforms to a certain schema (which `sails.io.js` ensures for you), it will be routed to the appropriate controller/action seamlessly.  Otherwise it should be ignored by the router but still handleable with regular socket.io semantics.  If you're seeing that log message in response to doing something like `io.socket.get("/foo")` in the client, then all is well.

sgress454:
I do like this, but as with most new features in Sails it should be implemented as an installable hook.  In this case it would be really easy to do--the whole `index.js` of the hook would just be:

```
module.exports = function(sails) {
  return {
    configure: function() {
      if (sails.config.http && sails.config.http.middleware && sails.config.http.middleware.order) {
        delete sails.config.http.middleware.order.startRequest;
        sails.config.http.middleware.order.unshift("requestTimer");
        sails.config.http.middleware.requestTimer = (process.env.NODE_ENV != 'production') && 
        function startRequestTimer(req, res, next) {
           ...timer function here...
        };
      }
    }
  };
};
```

Publish that as `sails-hook-hires-request-timer` (or whatever), with the format specified in the [docs for installable hooks](http://sailsjs.org/documentation/concepts/extending-sails/hooks/installable-hooks), and then anyone can do `npm install sails-hook-hires-request-timer` to get the new fancier timer!  This would allow you to make it more configurable, too--e.g. if a user wanted to be able to change the logging level from "verbose" to "silly" or something.

If you end up doing it this way, I'll see about putting a note in with the v0.12 release notes about your hook.

PR #2625, opened 2015-02-07
disable express' built in `x-powered-by` header

boogerlad:
sails sets it anyways. If someone disables sails' middleware for this header, they probably don't want express' either.

### DIFF
--- a/lib/hooks/http/initialize.js
+++ b/lib/hooks/http/initialize.js
@@ -28,7 +28,7 @@ module.exports = function(sails) {
         // Required to be here due to dynamic NODE_ENV settings via command line args
         // (i.e. if we `require` this above w/ everything else, the NODE_ENV might not be set properly yet)
         var express = require('express');
-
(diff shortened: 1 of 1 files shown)


loicsaintroch:
I am not sure to understand the goal here... Can you be more specific please?

boogerlad:
https://github.com/balderdashy/sails/blob/master/lib/hooks/http/middleware/defaults.js#L191 shows that sails sets it by itself. If someone wanted to disable the 'poweredBy' middleware, then it would fall back onto the express default, which is `x-powered-by: express`. I'm pretty sure that's unintended behavior(for me it is). If someone disables that middleware, then there should be no x-powered-byheader at all.

rachaelshaw:
@loicsaintroch @boogerlad thanks guys

loicsaintroch:
Oh ok I misunderstood what was the goal here @boogerlad. Thanks a lot.

PR #2633, opened 2015-02-10
[Patch] Blueprints actionUtil::parsePk also checks params for where.id.

marnusw:
Blueprints actionUtil::parsePk does a thorough check of most places where an id may be provided, but doesn't inspect the `where` property on the `params` which is where the `where` object is retrieved for `find` methods. If there isn't a specific reason for not looking in this location I'd suggest adding it. This will allow a `delete` from the socket client with the id passed on the options rather than the URL.

### DIFF
--- a/lib/hooks/blueprints/actionUtil.js
+++ b/lib/hooks/blueprints/actionUtil.js
@@ -115,6 +115,15 @@ module.exports = {
 
     var pk = req.options.id || (req.options.where && req.options.where.id) || req.param('id');
 
+    if (!pk) {
+      var where = req.param('where');
+      // If `where` parameter is a string, try to interpret it as JSON
+      if (_.isString(where)) {
+        where = tryToParseJSON(where);
+      }
+      pk = where && where.id;
+    }
+
     // TODO: make this smarter...
     // (e.g. look for actual primary key of model and look for it
     //  in the absence of `id`.)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### master
 
+* [ENHANCEMENT] Blueprints actionUtil::parsePk also checks params for where.id
 * [ENHANCEMENT] Support JSON sorting syntax in blueprints [#2449](https://github.com/balderdashy/sails/issues/2449)
 * [ENHANCEMENT] Support private modules as hooks [#3022](https://github.com/balderdashy/sails/issues/3022)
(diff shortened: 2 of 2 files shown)


mikermcneil:
@marnusw cool, thanks!  Would you mind rebasing so it can be automatically merged?

marnusw:
@mikermcneil, Done.

marnusw:
I see the Travis build has failed, but I'm not sure that it's my fault. If you can confirm with a quick glance great, otherwise I'll investigate this later.

CWyrtzen:
@marnusw have you had a chance to revisit this?

marnusw:
@CWyrtzen no I haven't. I tried running the sails tests locally once and there were problems even on the master branch and I didn't have time to investigate further. I'll see if I can get round to it in the near future though.

marnusw:
@mikermcneil, after rebasing to the latest master this PR passes the tests now without any additional changes. I guess when I did it last time there was just a temporary issue on the master branch or the failed test was a fluke. Either way, this should be ready to merge now?

lib/hooks/blueprints/actionUtil.js
```diff
@@ -115,6 +115,15 @@ module.exports = {
 
     var pk = req.options.id || (req.options.where && req.options.where.id) || req.param('id');
```
goliatone: Why are we using a hardcoded key name of **id**? In general sails assumes the pk would be named "id" but that is not always the case. Should `Model.primaryKey` be used instead?

marnusw:
@goliatone, I think this comment blow my changes might have reference to your query:
```javascript
    // TODO: make this smarter...
    // (e.g. look for actual primary key of model and look for it
    //  in the absence of `id`.)
    // See coercePK for reference (although be aware it is not currently in use)
```
It looks like it should work the way you suggest, but never has. I can easily include this change in this PR if that is the desired operation here.

@mikermcneil Would you like me to add the changes mentioned above?

Upon doing so I will also rebase this again, since I see new merge conflicts have arisen.

mikermcneil:
@marnusw that would be awesome-- apologies for the delay.  I'd consider both of these changes patches, since this is definitely the intended behavior.  All I'd ask is that you add a test that fails before your patch as part of this PR.  Thanks so much!

marnusw:
I haven't been using Sails in ages. If this PR is still of interest for the project please clone my branch and continue from there. I have also enabled edits from maintainers on this PR if that would be easier.

PR #2644, opened 2015-02-11
Remove unused dependency variable

igncp:
It seems like this dependency variable is unused since commit 3b77489c . Is this intentional or can it be removed?


rachaelshaw:
@igncp :+1:

PR #2647, opened 2015-02-12
Fix #2636 RangeError: Maximum call stack size exceeded on PubSub hook

### LINKED ISSUE #2636: PubSub triggers [RangeError: Maximum call stack size exceeded] in 0.11.0-rc10
I've upgraded my sails application to 0.11.0-rc10 today, and it works perfectly, with a very smooth update.

However it seems that there is a small issue with the upgrade to socket.io 1.0 because when i try to manually do a "publishCreate" on a collection without specifying which socket to omit from the socket message, it triggers a [RangeError: Maximum call stack size exceeded] error.

From my debugs, it triggers into the new socket hook, when it calls

```javascript
// Otherwise broadcast to everyone in the room.
app.io.sockets.in(roomName).emit(eventName, data);
```

Of course if i specify a socket to omit, it works because it calls:

```javascript
socketToOmit.broadcast.to(roomName).emit(eventName, data);
```

Maybe i'm doing something wrong? My call is:

```javascript
MyCollection.publishCreate(myInstance);
```

Thanks!

nanopx:
Make sure the object (i.e Model) is clean in "publishUpdate", "message" methods.

For the "publishCreate" method, it seems like the code below is avoiding the error.
```JavaScript
var legacyData = _.cloneDeep({
  verb: 'create',
  data: values,
  model: self.identity,
  id: values[this.primaryKey]
});
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### master
 
+* [BUGFIX] Fix [RangeError: Maximum call stack size exceeded] error in PubSub hook
 * [ENHANCEMENT] Support layout for Ractive template engine
 
 
--- a/lib/hooks/pubsub/index.js
+++ b/lib/hooks/pubsub/index.js
@@ -168,6 +168,10 @@ module.exports = function(sails) {
           var id = record[this.primaryKey] || record;
           // Get the socket room to publish to
           var room = this.room(id, "message");
+
+          // Ensure that we're working with a clean, unencumbered object
+          data = _.cloneDeep(data);
+
           // Create the payload
           var payload = {
             verb: "messaged",
@@ -537,7 +541,7 @@ module.exports = function(sails) {
         options = options || {};
 
         // Ensure that we're working with a clean, unencumbered object
-        changes = _.clone(changes);
+        changes = _.cloneDeep(changes);
 
         // Enforce valid usage
         var validId = _.isString(id) || _.isFinite(id);


erkstruwe:
Does this affect all publishXXX functions?
Does it respect custom .toJSON() functions?
If yes, then this is a good one.

My current alternative is in hooks/pubusb/index.js
```
broadcast: function(roomName, eventName, data, socketToOmit) {
  if (data.previous && _.isFunction(data.previous.toJSON)) {
    data.previous = _.clone(data.previous.toJSON(), true);
  }
  sails.sockets.broadcast(roomName, eventName, data, socketToOmit);
},
```

Skullbock:
@nanopx are you sure this takes toJSON() into account?

mikermcneil:
@particlebanana we were just looking at something similar in WL- what's the status on that?

@erkstruwe @Skullbock the `.toJSON()` you're suggesting is the right approach; although it's potentially a breaking change for apps that call toJSON manually.  If we can get tests in for this in the applicable parts, I'd like to bring it in for 0.12.

nanopx:
@erkstruwe @Skullbock this doesn't cover the ```.toJSON()``` method. It simply avoids the error.
I wasn't sure if implementing the custom ```.toJSON()``` method inside the hook was the best approach.

@erkstruwe Does your approach cover up the ```Model.message()``` method?
What if the second argument of the ```Model.message()``` was a model?

mikermcneil:
Although I'm pretty sure it won't solve every case (I just ∆ with this same issue in [rttc](https://github.com/node-machine/rttc/blob/a7fbcb5c4129796a071080d78a58b713a6bc3de6/lib/helpers/validate-recursive.js#L110)) it will help, so I'm merging

mikermcneil:
@nanopx @Skullbock @erkstruwe thanks y'all.

What would the desired behavior be?  Should things be collapsed into `[circular]`?  Could count up and use a maxhops approach like in `util.inspect`.  There's probably a module that does that on npm (in fact `json-stringify-safe` might do it already, haven't dug into the code yet)

Skullbock:
Hi @mikermcneil thanks for taking the time to look into this. I would expect that if The object is a model, then it's called toJson on it and then the data is broadcasted. In every other case I would just clonedeep th object passed

You idea of the max hops is great in the second case to avoid possible circular references

PR #2663, opened 2015-02-17
Enable `sails deploy`: Deploy Sails Apps within Minutes

felixrieseberg:
A Pull Request with :heart: from Microsoft! Crazy, I know. But hear me out: About a week ago, @mikermcneil, @sedouard and myself talked about how cool it would be if you could just run a command like `sails deploy` and sails would automatically create and deploy a website with the current sails app. We got some pizza and built that thing (Thanks to Mike for coming up to San Fran).

**TL;DR Version**
Run `sails deploy` and watch sails create an Azure Website and deploy your app. You might not have an Azure account - the script does some basic guidance to help you set one up, but this is really cool once you already have one. 

**Costs**
Azure has free accounts & up to ten free websites per account. The created websites are automatically set to be free. Nothing speaks against using the existing code to implement other providers.

**Full Feature List**
We first built `machinepack-azure`, which is capable of registering and setting up your Azure account, as well as creating a website, uploading files, uploading bash or powershell scripts, exposing said scripts via REST, calling saids scripts. Combined, it gives us the power to do pretty much anything.

The logical flow is as follows:
1) Check if an Azure account is setup and registered in the current `env`
2) Check if a website with the given name already exists in the current account
3) If not, one is created
4) The sails app is zipped up
5) The zip package is uploaded to the created website
6) A cleanup/unpack/install script is uploaded and registered and exposed
7) The script is called and the output piped through to the local console

PR #2667, opened 2015-02-18
Added ability to create custom sails commands

brooklynDev:
Currently `sails` does not have the ability to create custom commands at the `sails` command line. This PR adds that ability with the following rules:

- A command hook needs to specify `sails: { isCommand: true}` in the `package.json`
- A command hook needs to export an object with 2 properties `command` and `execute`.
  - the `command` property should be a string representing the command name
  - the `execute` property should be a function that will be called with the additional arguments.

For example a module like this will add an `azure` command that can be used like this `sails azure deploy`
```javascript
module.exports = {
  command: 'azure',
   execute: function(args) {
     console.log(args) // will log ['deploy']
  }
}
```

The motivation behind this PR was https://github.com/balderdashy/sails/pull/2663. The way that is currently set up is that if you do `sails deploy` it defaults to publishing to Azure. I think it would be a much nicer experience if the command would be `sails azure deploy` and in the future perhaps a `sails heroku deploy` etc. Additionally, this would allow the `azure` functionality to be wrapped up into it's own (opt-in) module.

### DIFF
--- a/bin/sails-command-hook.js
+++ b/bin/sails-command-hook.js
@@ -1,0 +1,38 @@
+var path = require('path');
+var buildDictionary = require('sails-build-dictionary');
+var _ = require('lodash');
+var program = require('commander');
+
+module.exports = function() {
+  var cliArguments = _.initial(arguments);
+  var commandName = cliArguments.shift();
+  var args = cliArguments;
+  buildDictionary.optional({
+    dirname: path.resolve('.', "node_modules"),
+    filter: /^(package\.json)$/,
+    depth: 2
+  }, function(err, packages) {
+    var objectKeys = Object.keys(packages);
+    for(var i = 0; i < objectKeys.length; i++){
+      var pkg = packages[objectKeys[i]];
+      var cmd = loadModule(pkg, commandName);
+      if(cmd){
+        cmd.execute(args);
+        return;
+      }
+    }
+    return program.usageMinusWildcard();
+  });
+};
+
+function loadModule(package, commandName){
+  if (package['package.json'] && package['package.json'].sails && package['package.json'].sails.isCommand) {
+    var module = require(path.resolve('.', 'node_modules', package['package.json'].name));
+    if(module.command){
+      program.command(module.command); //add to list of help commands
+      if (module.command === commandName) {
+        return module;
+      }
+    }
+  }
+}
--- a/bin/sails.js
+++ b/bin/sails.js
@@ -12,6 +12,7 @@ var NOOP = function() {};
 
 
 
+
 program
   .version(package.version, '-v, --version');
 
@@ -151,7 +152,7 @@ cmd.action(program.usageMinusWildcard);
 // Mask the '*' in `help`.
 program
   .command('*')
-  .action(program.usageMinusWildcard);
+  .action(require('./sails-command-hook.js'));
 
 
 


tarlepp:
Uuh, this seems nice.

mikermcneil:
@brooklynDev awesome thanks!  will merge simultaneous to the main azure PR.

What I'm thinking is `sails deploy` can use this sort of custom command integration, and then look to something in your project's .sailsrc or package.json file to determine which deployment strategy to use (azure/ec2/heroku/etc)

tarlepp:
or just ```sails deploy [something]``` where something is azure, heroku, or something that you have build on your own.

This is because users may have to deploy their apps to multiple places, eg stable, rc and dev.

tarlepp:
and imho ```sails deploy [something]``` should make some 'rules' for actual deployment implementation, so users can make own ones easily with proper guidelines.

tikotzky:
@mikermcneil I think that the `azure` integration should be bundled into its own module which would expose an `azure` command. For example to deploy you would run `sails azure deploy`. The `azure` command might have other subcommands other than `deploy` available and you would run those as `sails azure [subcommand]`

As for `sails deploy` that should be a hook that is defined in the projects `package.json` (or `.sailsrc`) the same way you would define which command to run for `npm test`.

So for example your `package.json` would look like ...
```
sails : { deploy: 'sails azure deploy' }
```
that would tell sails to run `sails azure deploy` when you run `sails deploy`.
This would also allow you to define a simple `deploy` script without need an entire module.
something like...
```
sails : { deploy: 'cp -r * /production/ ' }
```

mikermcneil:
@tikotzky that sounds perfect- I'd say let's put it in the .sailsrc file (since it allows for setting cross-project global overrides if you place a .sailsrc in your $HOME directory)

So I took a pretty thorough pass at speccing out how it ought to work (see below)

@tikotzky @brooklynDev what do you think?  Would you be able to knock this out? (i.e. the part that touches sails core) I can take care of setting up the `sails deploy` and `sails-deploy-azure` modules.

Or if anyone else is interested in getting involved on this effort, please let me know here (or on twitter @mikermcneil.)
Thanks!


> #### Contents
>
> 1. Custom commands in Sails
> 2. `sails deploy`
>


## Custom commands in Sails

##### example `.sailsrc` file

```json
{
  "commands": {
    "deploy": "sails-deploy"
  },
  "deploy": {
    "module": "sails-deploy-azure"
  }
}
```

##### In Sails core...

> (in `bin/sails.js`)

When the CLI is run as `sails deploy`, since this is an unrecognized command, Sails should load base config using the `rc` module (see `bin/sails-generate.js` for an example).  This will load any relevant config from the .sailsrc file(s) which apply to the current app.

Assuming we have a .sailsrc like the one above, there will be new `commands` and `deploy` namespaces in the app's configuration object.  So we'd be able to access:

```js
configObjectReturnedByRC.deploy.module       // => 'sails-deploy-azure'
configObjectReturnedByRC.commands.deploy   // => 'sails-deploy'
```

Sails core should then check to see if `commands` exist, and if so, loop through each one and see if there is a command matching what the user is trying to do (e.g. "deploy").  If no match is found, cli does exactly what it does today (show usage instructions).  But if a match _is_ found, then the CLI requires the configured module ("sails-deploy") from the context of the current sails project's "appPath" (`configObjectReturnedByRC.appPath`).  For example, something like this (warning: I did not run or double-check this code):

```js
var configObjectReturnedByRC = require('rc')('sails');
var matchingCommand = _.find(configObjectReturnedByRC.commands || [], function (val, key){});
if (!matchingCommand) {
  /*... handle this case exactly as happens currently, then return early */
  return;
}
var pathToRequire = require('path').resolve(configObjectReturnedByRC.appPath||process.cwd(), 'node_modules', matchingCommand);
var customCmd;
try {
  customCmd = require(pathToRequire);
}
catch (e){
  // check if this is a MODULE_NOT_FOUND error (look at e.type or e.code, can't remember)
  // if not, log an error message explaining that the custom command module referenced in one of the user's sailsrc files has an issue (then dump the original error stack and bail out/return early)
  // if so, log a nice error message explaining the command the user typed was recognized as a custom command defined in one of their sailsrc files, but it could not be required from the specified path.  Then bail out (return early)
}

// Finally, if that all worked, run the custom command.
customCmd({
  config: configObjectReturnedByRC
}, function (err){
  if (err) {
    // log msg explaining deployment failed, then log the original error stack  for details.
    return;
  }
  // log msg explaining deployment was successful.
});
```

Note that this approach means users will need to `npm install sails-noop --save` in their project in order to run `sails noop` on the command line.


##### Custom command packages

Custom sails commands are npm packages that export a function w/ two arguments- `options` and `cb` (following standard node conventions).

For example, here's a very simple no-op command  as an example (`sails-noop`).  Aside from its package.json , it has only one file: `index.js`.

```js
/**
 * Module dependences
*/
// ...


/**
 * Usage:
 * `$ sails noop`
 */
module.exports = function noopCmd (options, cb) {
  // `options.config` is provided with the raw config that Sails core gathered by running `rc`.
  return cb();
};
```


## `sails deploy`

`sails deploy` can be implemented as a custom command as specced out above.

##### the `sails-deploy` package

The job of the `sails-deploy` package is to  `require()` the configured strategy, then run it.  Eventually it can end up taking responsibility for generic shared logic between deployment strategies (e.g. talking to github, pulling/pushing/tagging, etc.)- but we won't know the right things to generalize until later when we have a few different strategies to look at.  For now, we'll just run the strategy.

This will be pretty similar to what we just did (which is also more or less how generators, adapters, and hooks work).  In the `deploy` object in the sailsrc file, the `module` property can be specified as any of the following:
1. an absolute path (if it starts with `/`)
2. a relative path from the sailsrc file (if it starts with `./` or `../`), or otherwise
3. an NPM package name (in which sails will attempt to `require()` the deployment strategy module from the node_modules folder within the sails project's `appPath`).


Note that this approach means a user will need to `npm install sails-deploy-noop --save` in her project and configure `"deploy": {"module": "sails-deploy-noop"}` in her sailsrc file.  Then she can run `sails deploy` on the command line and `sails-deploy-noop` will be used automatically.


##### `sails-deploy-*` strategies

The simplest no-op deploy strategy (e.g. `sails-deploy-noop`) is simply an npm package with a package.json and an `index.js` file:


```js
/**
 * Module dependences
*/
// ...

module.exports = function sailsDeployNoop(options, cb) {
  // `options.config` is provided with the raw config that Sails core gathered by running `rc`.
  return cb();
};
```

Again, we can get fancier here eventually as far as what happens in `sails-deploy` itself, but for now, a simple function is the right approach.



Let me know if I missed anything, and thanks!

mikermcneil:
here's an early stubbed out version of the generic `sails-deploy` module https://github.com/mikermcneil/sails-deploy

https://www.npmjs.com/package/sails-deploy

mikermcneil:
and here's the same sort of stubbed-out early version of sails-deploy-azure:

https://www.npmjs.com/package/sails-deploy-azure
https://github.com/mikermcneil/sails-deploy-azure

I went ahead and put both modules on npm so they can be used when testing the remaining work in core

tikotzky:
@mikermcneil I just reviewed quickly, one thing that stands out is...

You recommended that the way to add a command would be to have a `commands` property in `.sailsrc` which would list out all custom commands. The main issue I see happening with that approach is...  
it requires the user to do 2 steps on both install and removal...
1. `npm install sails-deploy`
2. add a `deploy` prop to ` the `commands` object in `.sailsrc`

I think the better path (at least for adding custom commands)  would be do something along the lines that this PR is suggesting, where the module that is adding the command has `sails: { isCommand: true }` in its `package.json`. This way you don't need to worry about keeping your `.sailsrc` in sync everytime you add/remove a command.

This does possible make it more difficult have globally installed commands, and also does increase the chance of conflict between different commands, but both of those issues can be worked out.

mikermcneil:
> I think the better path (at least for adding custom commands) would be do something along the lines that this PR is suggesting, where the module that is adding the command has sails: { isCommand: true } in its package.json. This way you don't need to worry about keeping your .sailsrc in sync everytime you add/remove a command.

@tikotzky I agree with you-- but if we start by using the `.sailsrc`, we can work out the kinks before adding too much magic, know what I mean?  It also gives folks the ability to be more explicit (I've learned the hard way about too much magic too fast)  We started off w/ hooks loading via sailsrc first before doing auto-loading for exactly the same reason.

That cool w/ you?

tarlepp:
There should be document to explain all these ```package.json``` custom options for ```sails``` property. Also there should be some guidelines how to use those.

mikermcneil:
@tarlepp agreed - https://github.com/balderdashy/sails/pull/2667#issuecomment-75474760 is the beginning of that -- currently only other core use case is for hooks.  I can put together a doc on that as soon as I hear back from @tikotzky and/or @brooklynDev whether they'll implement this

tikotzky:
@mikermcneil sorry, I really would love to be able to work on this, it looks like it would be a fun project to hack on. It's just that right now my hands are pretty full and I wouldn't feel comfortable committing to getting this done.

If this is still waiting to get implemented when I have some time freed up I would be happy to tackle it.

mikermcneil:
@tikotzky thanks for the heads up-- I'll keep hacking on it as time permits.  I've been doing some overall maintenance in core first- if you beat me to it, would be awesome :)

mikermcneil:
Closing this for now to keep tidy.

@all- please feel free to open a new PR using the [spec discussed here](http://michaelmcneil.com/post/114677471702/spec-for-custom-commands-deployment-strategies)

Thanks for the great discussion!

PR #2692, opened 2015-02-25
Properly add/cleanup process event listeners.

haavardw:
Causes issues when using sails in combination with test frameworks that
also adds process event listeners (e.g cucumber).

Don't use process.once as this replaces any existing listeners.
Properly remove listener using process.removeListener which doesn't
remove other listeners than our own.

This also solves the problem of doing multiple sails.lift/sails.lower (typically 
in a test framework) which would give node warnings like this: 
(node) warning: possible EventEmitter memory leak detected. 11 exit listeners
 added. Use emitter.setMaxListeners() to increase limit.

### DIFF
--- a/lib/app/lower.js
+++ b/lib/app/lower.js
@@ -111,6 +111,12 @@ module.exports = function lower(cb) {
           sails.removeAllListeners(key);
         }
 
+        var listeners = sails._processListeners;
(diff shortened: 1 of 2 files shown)


haavardw:
Didn't pass travis

PR #2693, opened 2015-02-25
Properly add/cleanup process event listeners.

haavardw:
Add and cleanup process event listeners properly.

Fixes warnings when doing lift/lower multiple times, e.g in before() and
after() functions in a test suite.
    
Tests:
  test/integration/lift.lower.test.js
  test/unit/app.lower.test.js

### DIFF
--- a/lib/app/lower.js
+++ b/lib/app/lower.js
@@ -110,6 +110,15 @@ module.exports = function lower(cb) {
           sails.removeAllListeners(eventName);
         });
 
+        var listeners = sails._processListeners;
+        if (listeners) {
+          process.removeListener('SIGUSR2', listeners.sigusr2);
+          process.removeListener('SIGINT', listeners.sigint);
+          process.removeListener('SIGTERM', listeners.sigterm);
+          process.removeListener('exit', listeners.exit);
+        }
+        sails._processListeners = null;
+
         // If `sails.config.process.removeAllListeners` is set, do that.
         if (sails.config && sails.config.process && sails.config.process.removeAllListeners) {
           process.removeAllListeners();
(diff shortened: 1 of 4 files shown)


sgress454:
> Don't use process.once as this replaces any existing listeners.

This is news to me!  `.once()` just means that the listener will only be fired once and then removed; it doesn't have any effect on other listeners attached to the event.

haavardw:
@sgress454 , you're right, my description was not correct. I've restored the use of .once() .on() to the original use, but kept the cleanup in app.lower(). I've added tests that demonstrates the warning that typicall arises in a test suite (test/integration/lift.lower.test.js) and a unit test that checks that we remove the eventlisteners on lower (test/unit/app.lower.test.js).

sgress454:
Hm, the tests fail when I run this--it looks like it might not be lowering correctly, because I get a
```
error: Trying to start server on port 1337 but can't...Something else is probably running on that port!
```
in the `app.lower` tests.

haavardw:
@sgress454 It fails in travis as well but not for me locally so I'm a bit puzzelled. Any clue on what kind of environment I need to test with? (I'm running node 0.12 on linux).

Also, it looks to me like the travis run fails during lift (not lower). Could it be that a previous test doesn't lower correctly?

washimimizuku:
This fails when Node is <0.12

mikermcneil:
@sgress454 @washimimizuku thanks y'all

@haavardw would you mind rebasing from master and giving this another go?  For a while this year, a PR was merged that added something to our npm test script that caused it to fail intermittently for reasons unrelated to the content of the PR (it was checking duplicate code).  It would definitely be worth it to give it a fresh run on the latest and see where it stands.

sgress454:
Works for me now on all Node versions, and passed Travis w/ Node 10.  I'd say it's good to go.

Issue #2550, opened 2015-01-11
Feature request: Support Access-Control-Expose-Headers


hinok:
It would be great if we could define custom headers that browsers are allowed to access in config/cors.js
It's all about `Access-Control-Expose-Headers`.

I forked and added very basic implementation.
If you want I can make PR.

config/cors.js looks now

```javascript
module.exports.cors = {

  // Stripped code for readability
  
  /***************************************************************************
  *                                                                          *
  * Which headers should be allowed for CORS requests? This is only used in  *
  * response to preflight requests.                                          *
  *                                                                          *
  ***************************************************************************/

  headers: 'content-type, Total-Count',

  /***************************************************************************
  *                                                                          *
  * Which headers in CORS response should be allowed for borwser? This is    *
  * only used in response to preflight requests.                             *
  *                                                                          *
  ***************************************************************************/

  exposeHeaders: 'Total-Count'

};
```

My solution: https://github.com/hinok/sails/commit/81ed85ec914c542afd8e9fcac3472c044ea11253

@gaprl mentioned that earlier: https://github.com/balderdashy/sails/issues/2146

loicsaintroch:
I like your idea @hinok. We would love you to write a pull request of course! Thank you for helping us making Sails a better framework.

aantipov:
+1

ArtworkAD:
@loicsaintroch https://github.com/balderdashy/sails/pull/2712

hinok:
@ArtworkAD Good, I didn't have time to write tests for that feature but I'm glad that someone made PR ;)

Issue #2542, opened 2015-01-08
Count blueprint for models


ghaiklor:
Hi guys.

**The problem**
Need make route that will be mapped to each model and realize get count of records of model.

**Why?**
Sometimes for mobile devs it can be helpful for getting total counts of some data. Like *how many users already registered*, etc...

**Where?**
I think it can be done with /model/count route or smth like that.

**Solution**
Add `count.js` file to `actions` folder with code like below
```javascript
var actionUtil = require('../actionUtil');

module.exports = function getCount (req, res) {
  var Model = actionUtil.parseModel(req),
      criteria = actionUtil.parseCriteria(req);

  Model.count(criteria, function(error, response) {
    if (error) return res.serverError(error);
    res.ok({count: response});
  });
};
```

What you think about this ?

ghaiklor:
BTW, I can make pull request for this if you think that this is the great idea.

randallmeeker:
https://gist.github.com/RWOverdijk/f64d2c4e5bab4922e861#file-count-js

ghaiklor:
@randallmeeker I mean, that will be great if this feature will be in Sails from the box. Why we need always write own blueprint if it can be integrated in Sails blueprints by default?

randallmeeker:
@ghaiklor I can't imagine there is any particular reason, other than it has not been done yet. I would submit your pull request. All my blueprints are usually re written anyway. Blueprints out of the box don't really work. I also have pagination in my blue prints which return counts already as well.

https://github.com/randallmeeker/SailsBluePrintActions/tree/master/pagination

ghaiklor:
@randallmeeker for me most of default blueprints is working as I expect and this fine. That's why I'm trying to ask guys implement this feature. But your variant great too. I should think about this.

animedbz16:
This request is fairly similar to one that I had proposed and worked with @randallmeeker:
https://github.com/balderdashy/sails/issues/1730

Basically ensuring that every API request would return a content-range HTTP header with the requests.

This way if you performed a search or had paginated results, you could at least have the content range requested and a total count returned in the same request.

At the point when I requested this, we were really testing out possibly switching over to Sails for one application, but ended up going in another direction for that app since we are pretty far along in development with it so I didn't have time to submit a PR

We using sails for a new app, but haven't had much time to go back and address that to submit a PR.

I think having this functionality would be great to be included in the default blueprints.

ghaiklor:
[...]

PR #2712, opened 2015-03-02
Allow clients to read custom response headers

ArtworkAD:
I was trying to read a Content-Range header in angularjs in order to build a proper pagination. But it simply was not visible to angular because the header was not exposed. We must allow developers to whitelist allowed headers with **Access-Control-Expose-Headers**.

https://github.com/balderdashy/sails/issues/2550
https://github.com/balderdashy/sails/issues/2542


sgress454:
[Merged](https://github.com/balderdashy/sails/commit/82b383f6eee7a8588d77802346f00ba2a40b9bd9).  Thanks!

sylvainlap:
The exposeHeaders attribute should be documented.

PR #2719, opened 2015-03-05
Fixed the Freenode link

lovemyjs:
Don't know if this is actually used over Gitter. But I thought I'd fix it regardless?


loicsaintroch:
Thank you @lovemyjs

PR #2732, opened 2015-03-10
Disabling CSRF via web-sockets accessed routes

juulderuysscher:
When you do a post to a route which is in the `sails.config.csrf.routesDisabled` string via a web-socket CSRF was not disabled.

### DIFF
--- a/lib/hooks/csrf/index.js
+++ b/lib/hooks/csrf/index.js
@@ -80,8 +80,12 @@ module.exports = function(sails) {
             } catch(err) {
               // Only attempt to handle invalid csrf tokens
               if (err.message != 'invalid csrf token') throw err;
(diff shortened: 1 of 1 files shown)


mikermcneil:
@juulderuysscher hmm I'm not sure I'm ok with doing this-- if we disable CSRF for sockets, how will we deal with [socket hijacking](http://sailsjs.org/#!/documentation/concepts/Security/SocketHijacking.html)?  [Here's how we normally suggest using CSRF over websockets](http://sailsjs.org/#!/documentation/concepts/Security/CSRF.html?q=using-ajax%2Fwebsockets) -- would you try that out?  Let me know if you run into difficulties.  Thanks!

PR #2756, opened 2015-03-17
req.param() is broken for boolean false params

jub3i:
for example if you posted data `{ isPirate: false }` and you run `req.param('isPirate')` from a controller, you'd get `undefined` returned as you didn't specify a `defaultValue` - so basically you couldn't post params that where === false. I think this edit captures the spirit of the OR guard used previously. If your happy with this, I would be ok with writing a test... if needed.

### DIFF
--- a/lib/router/req.js
+++ b/lib/router/req.js
@@ -94,7 +94,12 @@ module.exports = function buildRequest (_req) {
 
       // Grab the value of the parameter from the appropriate place
       // and return it
-      return params[paramName] || defaultValue;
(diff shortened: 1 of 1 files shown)


jub3i:
if anyone is interested a temporary workaround for this: instead of using `req.param('myParamName')` you would use `req.allParams().myParamName`

Enkows:
+1

mikermcneil:
Thanks @jub3i!

PR #2765, opened 2015-03-20
maxlen 80 added

juanpabloaj:


tgirardi:
I believe that what @juanpabloaj really wants to do is to avoid `sails new`, `sails generate` and other similar tools from generating files that are wider than 80 columns. IMHO that is a **good idea**.

In mi case, I use `maxlen: 80` as a global jshint configuration value (~/.jshintrc), so, after running `sails new` I always get a lot of "Line is too long" errors, from running jshint afterwards. At that point I have to choose between changing all the generated files or configuring a greater maxlen value for that particular application (which I don't like too much... 80 columns limit is a very common standard and I would like to use it also in my sails apps).

There is one thing to note though: `maxlen`has been **deprecated** for jshint, in favour of using JSCS' code style directives instead.

tjwebb:
maxlen 80 is pretty restrictive these days where my monitor has a kazillion pixels, and is not a standard of any kind: http://programmers.stackexchange.com/a/148678. 

I think maxlen should not be infinity, but 80 is too small. I'll wait for more input from the community on this.

tgirardi:
Yes, today screens sizes and pixel count make the 80 maxlen seem a little
restrictive, but there are cases in which it's very useful to have that
limit:

* For merging and conflict solving (the screen is usually divided by three
and a maxlen of 80 usually works great with Full HD screens in these cases)
* For side-to-side diffs and file comparing
* If you want to use your screen in vertical orientation (90º rotation),
which is handy for files with a lot of lines
* When you want to stack more than one application side-by-side (for
example, you editor and your browser)
* For browser's development tools that are docked to the main window and
aligned to the left or right side of it.
* For presentations where you want to use a big font size or if you are
using a datashow with a low resolution.

On 27 March 2015 at 21:10, Travis Webb <notifications@github.com> wrote:

> maxlen 80 is pretty restrictive these days where my monitor has a
> kazillion pixels, and is not a standard of any kind:
> http://programmers.stackexchange.com/a/148678.
>
> I think maxlen should not be infinity, but 80 is too small. I'll wait for
> more input from the community on this.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/balderdashy/sails/pull/2765#issuecomment-87128245>.
>



-- 
Tomás Girardi Julio
Gerente General
Turbomarket S.A.

Ingeniero Civil Electrónico, UTFSM
Mención Computadores - Submención Gestión

tjwebb:
I think there should be *some* limit, but 80 is too small.

sgress454:
Yeah, I'm hesitant to merge something that will make folks' code editors blow up in a blaze of "line too long" errors when they try to make a contribution to Sails (or when they open up a newly-generated Sails app!).  Adding this limit here also won't affect what the generators create.  

We've been historically slow to adopt new coding standards unless there was a very clear benefit coupled with large demand.  The best way to clear that bar is to create a discussion in our [Google Group](https://groups.google.com/forum/#!forum/sailsjs) and post another issue here when there's some consensus around the topic.  Open source democracy FTW!

PR #2770, opened 2015-03-21
make default responses work when views hook disabled

Enkows:
`res.guessView` and `res.view` is always `undefined` when views hook is disabled.

### DIFF
--- a/lib/hooks/responses/defaults/badRequest.js
+++ b/lib/hooks/responses/defaults/badRequest.js
@@ -21,6 +21,11 @@ module.exports = function badRequest(data, options) {
   var req = this.req;
   var res = this.res;
   var sails = req._sails;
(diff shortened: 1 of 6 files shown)


sgress454:
Hi @Enkows, this looks good.  Would love to see tests for this but it looks like we don't have tests for the responses hook in general, so I'll see about that.

sgress454:
Sorry for the delay.  Ended up going with a somewhat simpler approach [here](https://github.com/balderdashy/sails/commit/82e75971f442fd9087b5dfb86d5482526568564c), and merging [this](https://github.com/balderdashy/sails-generate-backend/commit/4645e3052c65483badf2e753e301b09fc394b4e7) into the generator that creates the responses.  Thanks for bringing this to our attention though!

PR #2771, opened 2015-03-22
Add node.js 0.12 and latest io.js for CI

listepo:


mikermcneil:
Thanks @listepo.  Hmm looks like tests fail on both new versions:

iojs:
https://travis-ci.org/balderdashy/sails/jobs/55336712
Looks like iojs breaks the `microtime` version we're using- can probably solve that by upgrading to the latest version of the module.  I think the [customFds option](https://github.com/joyent/node/wiki/API-changes-between-v0.10-and-v0.12#child-process) stuff is a red herring, but I could be wrong.



0.12:
https://travis-ci.org/balderdashy/sails/jobs/55336711
While it's not in the [changelog](https://github.com/joyent/node/wiki/API-changes-between-v0.10-and-v0.12), it looks like the upgrade to node v0.12 changes the HTTP methods exposed via cors; at least in the dependency we're using.  
Here's the broken test: https://github.com/balderdashy/sails/blob/master/test/integration/hook.cors_csrf.test.js#L91

We're expecting:
```
CHECKOUT,CONNECT,COPY,DELETE,GET,HEAD,LOCK,M-SEARCH,MERGE,MKACTIVITY,MKCOL,MOVE,NOTIFY,PATCH,POST,PROPFIND,PROPPATCH,PURGE,PUT,REPORT,SEARCH,SUBSCRIBE,TRACE,UNLOCK,UNSUBSCRIBE
```

But we're getting:
```
CHECKOUT,CONNECT,COPY,DELETE,GET,HEAD,LOCK,M-SEARCH,MERGE,MKACTIVITY,MKCOL,MOVE,NOTIFY,PATCH,POST,PROPFIND,PROPPATCH,PURGE,PUT,REPORT,SEARCH,SUBSCRIBE,TRACE,UNLOCK,UNSUBSCRIBE
```

mikermcneil:
@listepo actually, ignore the second part of that-- it's something else.  Probably a change in timing of execution-- looks like the test has a typo

mikermcneil:
@listepo thanks again for submitting this.

I just [removed iojs as a build target](https://github.com/balderdashy/sails/commit/767c6920bc4e5fe7af0534c9f9d67eb5e3449b6a) for now until the microtime issue is sorted.  If this is important for you or anyone reading this, please try upgrading the dep locally and running the tests.  If you're stuck, let me know.  Thanks!

listepo:
Ok, I will try to solve the problem

listepo:
@mikermcneil I fix microtime https://github.com/wadey/node-microtime/pull/25

PR #2778, opened 2015-03-23
added missing ! on link for extending sails doc

luislobo:
added missing ! on the link for extending sails doc


tjwebb:
Nice catch, thanks @luislobo

PR #2784, opened 2015-03-25
give ability for hooks to modify loaded models before orm is built

ynk:
According to [orm/index.js#L73-L78](https://github.com/balderdashy/sails/blob/master/lib/hooks/orm/index.js#L73-L78), a hook could be able to add or alter models by modifying `sails.models` after event `hook:orm:loaded` and call for a reload.

The line responsible for such a behavior is [this one](https://github.com/balderdashy/sails/blob/master/lib/hooks/orm/load-user-modules.js#L25) `sails.models = _.extend(sails.models || {}, modules);` but as lodash extend is not a deep copy feature, it only provides ability to *add* new models but not to *alter* existing ones.

Because each model in `sails.models` is populated by bunch of stuff related to waterline, it seems better not to try to implement deep copy, but i rather propose an event that would trigger just after that merge, giving the ability to a hook to alter `sails.models`. It should be before any other step in the process so it will go through a normal flow of normalization and etc...

(edit: [usecase](https://github.com/ynk/sails-hook-gravatar/tree/fix/methodAttachedToInstance))

### DIFF
--- a/lib/hooks/orm/index.js
+++ b/lib/hooks/orm/index.js
@@ -85,6 +85,7 @@ module.exports = function(sails) {
         // Load model and adapter definitions defined in the project
         _loadModules: function (next) {
           loadAppModelsAndAdapters(next);
+          sails.emit('hook:orm:models:loaded');
         },
 
         // Load any adapters for connections with "forceLoadAdapter"


blah238:
:+1:

tjwebb:
This is already possible; an example is here:

- https://github.com/tjwebb/sails-permissions/blob/master/api/hooks/sails-permissions.js#L29
- https://github.com/tjwebb/sails-permissions/blob/master/api/hooks/sails-permissions.js#L84-L100

ynk:
yeah i looked at your repo before :) it appeared to me that the provided documentation wasn't correct, and bugfix was too complex so this little event was the best compromize.

sgress454:
I get what you're going for here, but putting an event emitter in that `async` block is not great practice.  You're counting on any event handlers for `hook:orm:models:loaded` to be synchronous, which you can't guarantee; otherwise, the ORM might continue initializing while some handler is working, and things could get merged into `sails.models` at the wrong time.

If the issue is that you want to modify Sails model definitions after the ORM loads, then I'd like to see a failing test where you attempt to do so using the technique that @tjwebb pointed out above: namely, adding/modifying attributes directly on the model.  If there's something that you can't already do using that technique, then we'll brainstorm solutions.

ynk:
You're right about the sync presomptions. Thanks for pointing that out :)

PR #2785, opened 2015-03-26
Binding policies functions to `sails` object

konstantinzolotarev:
Binded all policies form `api/policy` folder to `sails.policies` object. 

This will be very helpful for hooks. Hooks will be able to add new custom policies into sails application.


Kikobeats:
An example of case of use: https://github.com/Kikobeats/sails-hook-jwt/blob/master/index.js#L52

sgress454:
@konstantinzolotarev @Kikobeats  what does this allow that you couldn't do by referencing `sails.hooks.policies.middleware`?

tjwebb:
While this is a "feature" in that it allows you to do something new, policies aren't really designed to be invoked directly. I think this offers too many ways to circumvent sails convention

Kikobeats:
@sgress454 really I didn't know how to do it. Can you put an example about it? If is possible have the same behavio, then is not necessary. Write a little documentation about it for future issues can be useful.

konstantinzolotarev:
I also didn't saw this solution. 
@sgress454 Thank you !

@Kikobeats here is example:
```javascript
if (!sails.hooks.policies.middleware) {
    sails.hooks.policies.middleware = {};
}
sails.hooks.policies.middleware.loadUser = function(req, res, next) {
    sails.log.debug('this policy exists !');
    next();
};
```

konstantinzolotarev:
@Kikobeats 
But one small note. Policies from config file `policies.js` are checking when hook `policies` is loading. So you have to do this operation before this hook will load. Or you will get this error:
```
error: Unknown policy, "%policyName%", referenced in `config.policies`.
error: Are you sure that policy exists?
error: It would be located at: `*****/api/policies/%policyName%.js`
```
But I didn't find any way to handle this except of changing hook load order. 
@sgress454 might be you could help us with it ?

konstantinzolotarev:
I'll close this PR and create a new one to solve this problem in few days.

PR #2792, opened 2015-03-28
Make sails compilable under io.js and pass the current test suite

serjrd:
Today I decided to give io.js (v1.6.2) a try.

My project seems to be working fine, so I decided to run an `npm test` to test sails.
The issues I found were really very minor:

- microtime devDep v0.6.0 didn't compile right under io.js, so I updated it to the latest version, which uses the same API
- the HTTP OPTIONS string seems to be alphabetically sorted now, so I updated the tests accordingly (the supertest one could be done better though)

Hope this will be at least a minor help/thank you for making such a great project :)

P.S.: Congratulations on getting 10000 stars! :)

### DIFF
--- a/test/integration/hook.cors_csrf.test.js
+++ b/test/integration/hook.cors_csrf.test.js
@@ -86,8 +86,9 @@ describe('CORS and CSRF ::', function() {
             url: 'test',
           }, function(err, response) {
             if (err) return done(new Error(err));
+            body = response.body.split(',').sort().join(',');
             assert.equal(response.statusCode, 200);
-            assert.equal(response.body, "GET,POST,PUT,HEAD,DELETE,TRACE,COPY,LOCK,MKCOL,MOVE,PURGE,PROPFIND,PROPPATCH,UNLOCK,REPORT,MKACTIVITY,CHECKOUT,MERGE,M-SEARCH,NOTIFY,SUBSCRIBE,UNSUBSCRIBE,PATCH,SEARCH,CONNECT", require('util').format('Unexpected HTTP methods:  "%s"', response.body));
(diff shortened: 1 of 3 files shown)


test/integration/hook.cors_csrf.test.js
```diff
@@ -86,8 +86,9 @@ describe('CORS and CSRF ::', function() {
             url: 'test',
           }, function(err, response) {
             if (err) return done(new Error(err));
+            body = response.body.split(',').sort().join(',');
             assert.equal(response.statusCode, 200);
-            assert.equal(response.body, "GET,POST,PUT,HEAD,DELETE,TRACE,COPY,LOCK,MKCOL,MOVE,PURGE,PROPFIND,PROPPATCH,UNLOCK,REPORT,MKACTIVITY,CHECKOUT,MERGE,M-SEARCH,NOTIFY,SUBSCRIBE,UNSUBSCRIBE,PATCH,SEARCH,CONNECT", require('util').format('Unexpected HTTP methods:  "%s"', response.body));
+            assert.equal(body, 'CHECKOUT,CONNECT,COPY,DELETE,GET,HEAD,LOCK,M-SEARCH,MERGE,MKACTIVITY,MKCOL,MOVE,NOTIFY,PATCH,POST,PROPFIND,PROPPATCH,PURGE,PUT,REPORT,SEARCH,SUBSCRIBE,TRACE,UNLOCK,UNSUBSCRIBE', require('util').format('Unexpected HTTP methods:  "%s"', response.body));
             done();
```
tjwebb: Even better would be to use `_.contains`. Now won't it fail for pre-iojs versions?

serjrd: No, it won't. Coz now it always sorts the list of methods before the assertion.
I just checked it on a regular node (v0.10.37) - all tests pass just fine.

As for the `_.contains` - I just don't see how exactly it'd fit here :)

tjwebb: Oh, I meant breaking up those CSVs into a list and checking existence, then order won't matter.

PR #2803, opened 2015-04-01
Update dependencies and solve problem on node 0.10 and iojs

listepo:


listepo:
Fix test /home/travis/build/balderdashy/sails/test/integration/hook.cors_csrf.test.js for node 0.10(express 3.20.2)

sgress454:
Hi @listepo can you go into a bit more detail as to what the issue is that this PR is solving?  It looks like you just updated all of the dependency versions, but it's not clear what the strategy was in doing so.

listepo:
Most modules contain fixes minor bugs and improved performance

listepo:
Upgrade express.js to 3.20.2 slove difference node 0.10 and 0.12 for hook.cors_csrf.test.js

listepo:
Considered good form to do the updates and update modules in the project more

listepo:
Solve problem sails.js on iojs see https://github.com/wadey/node-microtime/pull/25 and https://github.com/balderdashy/sails/pull/2771#issuecomment-88531136

listepo:
Ping

listepo:
Fix "npm WARN package.json Dependency 'colors' exists in both dependencies and devDependencies, using 'colors@~0.6.2' from dependencies" too

listepo:
Ping

sgress454:
Thanks for the reminder @listepo.  This took a little longer to get to b/c some of the dependencies were bumped considerably.  Our test coverage at this point is very good, but still not 100%, so I had to double-check some things manually (like handlebars support).  I also removed the Code Climate badge; if you think it's valuable you can open a separate PR for that so we can have a discussion around it and you can make your case.

listepo:
@sgress454, thanks.

PR #2816, opened 2015-04-06
Grammatical modifications

AndrewHendrie:
Cleaned up the grammar a bit (mostly removing unnecessary comma's). Clarified the section on line 56 about Waterline.


sgress454:
Thanks for the submission Andrew, but I don't think this is a substantial improvement to the document.  If you find parts that are actually unclear or *un*grammatical, please do submit fixes, but double-check first that they don't introduce new errors!

PR #2821, opened 2015-04-08
Fix for unknown policy issue if policy was added from another hook.

konstantinzolotarev:
Description form PR #2785: 

But one small note. Policies from config file `policies.js` are checking when hook `policies` is loading. So you have to do this operation before this hook will load. Or you will get this error:
```
error: Unknown policy, "%policyName%", referenced in `config.policies`.
error: Are you sure that policy exists?
error: It would be located at: `*****/api/policies/%policyName%.js`
```
But I didn't find any way to handle this except of changing hook load order.


lib/hooks/policies/index.js
```diff
@@ -89,6 +86,9 @@ module.exports = function(sails) {
 
 		bindPolicies: function() {
 
+      // Build / normalize policy config
+      this.mapping = this.buildPolicyMap();
```
tjwebb: why is the indentation way off here?

konstantinzolotarev: Because here it allow to check policies existence only after all hooks are loaded.
If it will be as now, some hooks wouldn't be able to add policies before this check and application will fail with error: 'no such policy found'

tjwebb:
@particlebanana @mikermcneil @sgress454 are you able to review this?

RWOverdijk:
What about making it possible to dynamically add policies? That question will come at some point anyway, and it seems like a more elegant approach to me.

tjwebb:
This is sort of possible by cramming them into sails.config.policies in a `hook.configure` function, but it's annoying and hard. 

I submitted a PR to make extending controllers somewhat easier: https://github.com/balderdashy/sails/pull/3083

I'll do something similar for policies when I have time. Unfortunately, merging arrays is somewhat more contentious than merging objects.

RWOverdijk:
@tjwebb Policies are flat arrays right? So merging those shouldn't be too tricky, just use concat. Besides, in an ideal world, policies bundles with hooks don't touch your main endpoints, only the ones they expose themselves.

Either way, this PR doesn't look horrible. I just thought it might be easier :)

tjwebb:
> So merging those shouldn't be too tricky, just use concat. 

Well of course. Except, you have to remove duplicates. And also what if you want to add a policy in between two other policies? It's not as simple as it seems...

RWOverdijk:
That last bit is simply doing a if (!Array.isArray(theThing)) { theThing = [theThing] }. But I can see your point. The sorting shouldn't matter too much though, should it? I mean, as long as they're grouped. I'll hit you up on le chattersphere

PR #2825, opened 2015-04-09
Fix issue #2484

### LINKED ISSUE #2484: Socket subscribe events do not call toJSON
I'm using sails v0.10.5, and subscribing on model events that have a toJSON seem to indicate that the toJSON is only called for blueprint update actions (see https://github.com/balderdashy/sails/blob/master/lib/hooks/blueprints/actions/update.js#L72 - this is the only spot I could find to do with pubsub that calls toJSON).
I'm pretty sure this isn't expected behavior, but maybe I'm wrong? Why would only updates send back a message with toJSON invoked, but nothing else does?

nickeddy:
- fixes #2484 

@tjwebb Hopefully this helps.


tjwebb:
thanks @nickeddy !

PR #2827, opened 2015-04-09
Pass configuration to sails deploy

felixrieseberg:
We're currently passing an empty configuration to the sails deploy module. Obviously, we need to pass the actual configuration object.


tjwebb:
Thanks @felixrieseberg!

PR #2866, opened 2015-04-23
Use the upstream consolidate.js instead of the copy-pasted version.

aclave1:
This pr uses the upstream version of consolidate.js from TJ H. @tjwebb and I discussed this the other night and we decided that adding a new function to consolidate.js every time a new rendering engine comes out is too much work when we can just bump the version on this. 

Info:

The original code wrapped a copy-paste from consolidate.js in a function. This function was to provide consolidate with a path to the user's current sails project's node_modules directory. I tested the upstream consolidate with sails and concluded that this is no longer needed. All tests pass and I've also manually verified by rendering some pages with sails. This is a crucial part of sails so I'd like someone else to verify that this works before merging in this PR.


sgress454:
Thanks for giving this a shot, Alex, but alas the reasons for the monkey-patched Consolidate are still valid.  The tests pass because they're using template engines that are bundled with Sails.  But try it with Swig, for example, and you'll see the issue.  

That being said, I'd love to see an alternate solution!  At the time we looked into ways of manipulating the require path so that [lines like this](https://github.com/tj/consolidate.js/blob/master/lib/consolidate.js#L379) would get the Swig installed locally to the Sails project.  Any ideas there would be welcome!

tjwebb:
> but alas the reasons for the monkey-patched Consolidate are still valid. 

Can you elaborate on this? We're trying to add support for the react engine, which is now included in consolidate. I don't want to monkey-patch stuff in forever.

sgress454:
@tjwebb try it with Swig!  Start a new project with the branch, do `npm install swig` in the project, set the `engine` in `config.views` to `swig`, and make a `views/homepage.swig` file.  You'll get:

```
Error: Cannot find module 'swig'
```

when you try to load the page, because Sails is trying to require Swig starting from the `node_modules` folder underneath the Sails install, *not* the local project.  We tried at the time to find a way to work around the issue without using a patched version of Consolidate, but eventually gave up.  That doesn't mean such a solution doesn't exist though.

tjwebb:
I believe you that it doesn't work, but I'm having trouble believing that this PR can't be salvaged with some npm path magic or something. e.g. tons of other modules are required from sails that are in the local `node_modules`. Like, I add `require('thing')` in a sails controller and it works great, after I npm install it. What is special here? Is it just that sails happens to be installed globally by default?

sgress454:
The difference is that the Sails controller file lives in the project, so when you do `require('thing')` from there, Node starts looking for `thing` in the *project's* node_modules folder instead of the *Sails* install's folder.  You're right in that this is only an issue when using the global Sails install--if you have Sails installed locally in your project's node_modules folder, it's fine, because Node fails to find Swig inside of `myproj/node_modules/sails/node_modules/swig`, then looks in `myproj/node_modules/swig` and finds it there.  But for the global install, when it crawls up the directory tree it never gets into the local project folder because it's in a completely different part of the filesystem.

sgress454:
(didn't actually mean to reopen just now but if you think you can somehow get it done with this PR, do it!)

tjwebb:
> if you have Sails installed locally in your project's node_modules folder, it's fine,

This sounds like the same issue where adapters can't be installed globally. I'd think everything might Just Work if we add `process.cwd()` to `NODE_PATH`... maybe we can actually get two things done here.

@aclave1 thoughts?

sgress454:
@tjwebb I don't think you can set NODE_PATH programmatically from within a running script...

aclave1:
@tjwebb @sgress454 

What we could do is either:

1. Submit a PR adding a property to consolidate like appBasePath
2. Fork and add this property

This way, any merges with the upstream consolidate don't have to be added manually via copy paste, we can do an easy merge with or if our pr gets accepted then we don't have to do anything.

aclave1:
@tjwebb @sgress454 looks like I had the same idea as mike [here](https://github.com/tj/consolidate.js/issues/135). A year later, an example of this was posted: [here](https://github.com/nymag/multiplex-templates/blob/master/index.js#L11).  Since consolidate checks to see if requires are cached [example](https://github.com/tj/consolidate.js/blob/master/lib/consolidate.js#L173) We can just require the view inside of sails with appBasePath and just tack it onto consolidate's requires object. When consolidate actually goes to render, it will check to see if its cached and everything should work.

aclave1:
Here's some code illustrating my idea:

```
//sails/lib/hooks/views/configure.js

fn = consolidate[engineName];
if(!consolidate.requires[engineName]){
 try{
   consolidate.requires[engineName] = require(appDependenciesPath+'/' + engineName);
 }catch(e){
   console.log(['Could not find module:',engineName,"in path:",appDependenciesPath].join(" "));
 }
}
```

tjwebb:
@aclave1 ah, good find. Yea looks like we're hitting an issue that's been around for awhile. This issue also occurs with adapters; adapters cannot be installed globally. Thanks for your willingness to help us solve this.

Do you have any thoughts on how you'd go about solving this globally, i.e. for both adapters, views, and whichever other things need to dynamically require stuff?

aclave1:
Hmm not sure. I looked into the way sails loads adapters and cant really find a good idea for where it does the searches. It appears that inside of hooks/moduleloader sails uses  a module called [sails-build-dictionary](https://github.com/balderdashy/sails-build-dictionary/blob/master/index.js) which uses a module called [include-all](https://www.npmjs.com/package/include-all). Here is the code that loads the adapters:
```
loadAdapters: function (cb) {
      buildDictionary.optional({
        dirname: sails.config.paths.adapters,
        filter: new RegExp("(.+Adapter)\.(" + sails.config.moduleloader.sourceExt.join('|') + ")$"),
        replaceExpr: /Adapter/,
        flattenDirectories: true
      }, bindToSails(cb));
    },
```

But I'm not sure where sails actually looks at your  config/connections, looks at each connection, grabs the module, and actually requires it.

tjwebb:
Hmm, also if this PR is so problematic, why do the unit tests pass... :/

aclave1:
Fixed the problem @sgress addressed in #2951. Closing this PR now

sgress454:
Looks great--merged.  Thanks errybody.

Issue #2670, opened 2015-02-19
"sails debug" on node v0.12.0 throwing error EADDRINUSE


MikeCostello:
Running Sails v0.11.0 and Node v0.12.0

When I run
```
sails new app; cd app
sails debug
```

Console logs...

```
error: Grunt :: Error: listen EADDRINUSE
    at exports._errnoException (util.js:746:11)
    at Agent.Server._listen2 (net.js:1129:14)
    at listen (net.js:1155:10)
    at Agent.Server.listen (net.js:1240:5)
    at Object.start (_debugger_agent.js:20:9)
    at startup (node.js:86:9)
    at node.js:814:3
```
No conflicting ports were listening prior to running sails.
Looking at travis.yml is only node v0.10.x supported currently?

Kikobeats:
at this moment sails pass the tests for 0.10.x, so I think that is not possible run your code under 0.12 because is not officially supported... 

(Although I do not think that error is the version of node).

tjwebb:
> error: Grunt :: Error: listen EADDRINUSE

There's only one way this happens. It's if your port is already being used by another program.

MikeCostello:
Agreed, but this is the output just prior to launching ```sails debug```
```
lsof -i -P | grep -i "listen"

redis-ser  678 mikecostello    4u  IPv6 0x704352d066f66a57      0t0  TCP *:6379 (LISTEN)
redis-ser  678 mikecostello    5u  IPv4 0x704352d06e87f7d7      0t0  TCP *:6379 (LISTEN)
postgres  1604 mikecostello    5u  IPv6 0x704352d066f68357      0t0  TCP localhost:5432 (LISTEN)
postgres  1604 mikecostello    6u  IPv4 0x704352d07074ad67      0t0  TCP localhost:5432 (LISTEN)
postgres  1604 mikecostello    7u  IPv6 0x704352d066f66f57      0t0  TCP localhost:5432 (LISTEN)
GitHub    8261 mikecostello   15u  IPv4 0x704352d07dc24d67      0t0  TCP localhost:25035 (LISTEN)
GitHub    8261 mikecostello   16u  IPv6 0x704352d066f67957      0t0  TCP localhost:25035 (LISTEN)
```

After running ```sails debug``` the debug port is opened and I can connect with node-inspector. Just the error is logged in the console.

Node v0.10 & Sails v0.11 works
Node v0.12 & Sails v0.10.5 same error

tjwebb:
[...]

PR #2888, opened 2015-04-29
explicitly set fork args before grunt hook

dbuentello:
The grunt process is being executed with its parents process arguments (i.e. `--debug`). This in turn causes a EADDRINUSE exception to occur.

Issue 
#2670 

Explanation
https://github.com/balderdashy/sails/issues/2670#issuecomment-92140615

### DIFF
--- a/lib/hooks/grunt/index.js
+++ b/lib/hooks/grunt/index.js
@@ -50,6 +50,11 @@ module.exports = function (sails) {
         taskName = '';
       }
 
+      var execArgs = process.execArgv.slice(0);
+      if(execArgs.indexOf('--debug') !== -1) {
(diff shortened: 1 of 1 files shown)


tjwebb:
You're zeroing out the execArgs. What if I want to include other arguments?

dbuentello:
Hmm good point. Considering this only affects the Grunt hook child process, how common is this?

tjwebb:
I'm not sure, but I don't want to fix one problem and create another :)

Can you just remove the `debug` argument from argv?

tjwebb:
@dbuentello do you have time to take another look at this?

dbuentello:
@tjwebb How does that look?

PR #2895, opened 2015-04-30
Feature Req(Backlog) for translating routes using i18n

Zaggen:
Assuming i have a multilingual site, that requires pages in **English** and **Spanish**, if i want **routes** to be on the appropriate language, then i'll need to manually create multiple routes to target the same controller. Like this.
**routes.js**
```javascript
module.exports.routes = {
   // Some routes
   'services': 'ServiceController.index',
   'servicios':  'ServiceController.index',
   'products':  'ProductController.index',
   'productos':  'ProductController.index', 
  // More routes
}
```
Now those are just four routes, but imagine now that you are supporting 5 languages and you have 30 base routes, you will have to manually enter 150 routes just for translation purposes, and it'll be really hard to maintain them.
So, what i'm proposing is a feature to use the i18n `__()` inside the routes keys, so i just have to add a single translatable route and sails will handle the rest. So here is how it should look:

```javascript
module.exports.routes = {
   // Some routes
   '__(services)': 'ServiceController.index',
   '__(products)':  'ProductController.index',
  // More routes
}
```
A lot cleaner, right?
You of course will need to provide the appropriate translations in the locales folders, but its easier to mange the routes this way.

I'll create a PR with a hook to add this functionality, but i'll need to modify  `router:bind`  in **sails/lib/hooks/http/initialize.js** like this

```javascript
   // Require i18nRoutes Hook
   // Some code ...
   // When Sails binds routes, bind them to the internal Express router
   sails.on('router:bind', function(route) {
       route = _.cloneDeep(route);
       var routes = [route];
       if(i18nRoutes.use){
          routes = i18nRoutes.getTranslations(route); // Returns an array with all translated routes
        }
        // TODO: Add support for error domains..?
        _.each(routes, function (route) {
           app[route.verb || 'all'](route.path, route.target);
         });
     });
```

Edit: I removed most of the code in this comment to place it on a proper pull request, and since this is a backlog feature, i'll just have the proposal here. Any advice is appreciated.


Zaggen:
I'll make this module into a sails hook and send a pull request soon. Any advice? What do you think of this feature?

Zaggen:
@jdcauley Any thoughts on this feature?

mikermcneil:
This is a very late response, @Zaggen; apologies.

I like the idea here, and I think it's a great feature.  I'm not 100% sure about the `__()` calls as a string expression, but I can't think of any cleaner way to do it either.  I believe making this change in core would involve more than changing router:bind, because it would also needs to change the way the `router:route` (for socket requests). It would also introduce a new kind of route address syntax, which would mean some changes to the documentation.

That said, I think a better approach here would be to take advantage of _something_ like the `route:typeUnknown` event in order to build this as a hook (or potentially to extend the core i18n hook- which I'd be more than down for after seeing it working in a standalone).  

Instead of `route:typeUnknown`, since it's the route _address_ rather than the target which cues the special behavior, I think you'll want to look into using your hook's `initialize` function to transform the explicit routes directly within `sails.config`.  To implement this in a standalone hook, you'll have to wait for the i18n hook to finish loading:

```js
sails.after('hook:i18n:loaded', function (){
  // find the set of keys (route addresses) in `sails.config.routes` which are like `__(..)`
  // then replace them inline in `sails.config.routes` with the internationalized versions of the route addresses
});
```

...so that you can use the methods it exposes.  Even waiting for i18n to finish loading, you'll have plenty of time to transform the configured explicit routes (`sails.config.routes`) in your initialize function because the Sails router does not actually bind explicit routes until all hooks have called the callbacks of their initialize functions.

Hope that gets you off to a good start-- and again, sorry for the long delay.  Please keep me posted on how this goes.  The best way to do that is, when you reach the point where you've got this working as a custom hook, submit a [proposal](https://github.com/balderdashy/sails/blob/master/CONTRIBUTING.md#v-proposing-features-and-enhancements).  I think this would be a great addition to the i18n hook.

Zaggen:
Hey @mikermcneil, thanks for your response, i know we all have a lot of work so don't mind about it. 
> Sails router does not actually bind explicit routes until all hooks have called the callbacks of their initialize functions

Knowing that its extremely helpful, i'll check into that as soon as i have time, but since its been a while, and i was needing this feature, i decided to implement it on a different way, by modifying the routes before they get exported, so i created an npm module to help with that (it has other purposes too), its still on development, but i'll paste a link here, and i'll like to know if you think that even with this module, will it be a nice feature to add to sails.js.

The module is called  [sails-navigator](https://www.npmjs.com/package/sails-navigator), and to keep with the same example above, it'll be something like this:
```js
// routes-translations.json (this could be a json file or js/cs object)
{
  "/services":  {"en": "/services", "es": "/servicios"},
  "/products":  {"en":  "/products", "es": "/products"}
}
```

```js
var navigator = require('sails-navigator').config({
  localizeRoute: ['es', 'en']
  localizedData: require('./locales/routes-translations')
})

module.exports = navigator(function(makeRoute){
  makeRoute('/services')
    .REST('index')

  makeRoute('/products')
    .REST('index')
});
```

Right now it only supports translation of restful routes, so it needs more work. I'll probably be using sails-navigator for my projects, but like i said if you guys think it'll be better for sails users to have this as a hook it might work on it, or in the other hand if you think its better to have this functionality as an external module like this, let me know that too.

mikermcneil:
> so i created an npm module to help with that (it has other purposes too), its still on development

Awesome :)


> I'll probably be using sails-navigator for my projects, but like i said if you guys think it'll be better for sails users to have this as a hook, then i might work on it, or in the other hand if you think its better to have this functionality as an external module like this, let me know that too.

@Zaggen I would say if it ain't broke, don't fix it.  The only risk of using higher-level abstractions to provide features like this is when/if they conflict with core or other community hooks.  But realistically this is a risk with any plugin system anyways, and the potential benefits of already-working support for this existing in _any form_ outside of core (for you and for everyone else with a similar use case in the community) far outweigh any potential compatibility issues down the road, IMO.

Anyway, this sounds like a perfectly valid approach to me-- in fact, when you've got this running within a production app and feel good about it, I think it'd be great to link to your project from the relevant section(s) of sails-docs (most obvious spot is probably the [conceptual section on internationalization](http://sailsjs.org/documentation/concepts/internationalization), with a close second being the [reference section on i18n config](http://sailsjs.org/documentation/reference/configuration/sails-config-i-18-n) )

Thanks for sharing the link!

Zaggen:
@mikermcneil I'm glad you liked, then i'll follow your suggestion, and keep working on that module. Regarding to the docs, when i have this module stable, should i do a pull request to the docs (i've done once to add some info on the unit testing section) or should i comment here and tag you, so you can update it?

As a side note, since you want me to link my module to the internationalization section, i'll add here another module just as relevant, [sails-routes-helpers-maker](https://www.npmjs.com/package/sails-routes-helpers-maker) it creates paths view helpers (rails style) by using  `config/routes` (created manually or by the sails-navigator) and the same translations file `routes-translations.json`.

So i'm guessing that i should notify you by commenting here when both of those modules are fully working and then we proceed with updating the docs (Let me know how we go with this part).
 :)

mikermcneil:
@Zaggen if you don't mind, would you create a new pr in sails-docs when they're ready? That way we can go ahead and close this, and @rachaelshaw, @brettimus, or I will be sure to see it. The best place to add on those pages is in the "notes" section (if there isn't one yet, feel free to create it-- conventionally we use an h3
with a list of bullets inside a block quote underneath)

Thanks!

Zaggen:
@mikermcneil Sure, no problem, i'll create a pr when they're ready ;)

mikermcneil:
@Zaggen thanks!

PR #2910, opened 2015-05-06
added "noBind" option to services to resolve #2791

### LINKED ISSUE #2791: Using Services modifies constructor functions
Example Code: https://bpaste.net/show/7c80367ce3a9

I know printing \_\_proto\_\_ isnt exactly best practices, but it serves my purpose here.  

When using a service to create, for example, an errors object, that constructs specific Errors for me.  In practice, I use this in bluebird promises `.catch()` to catch specific errors, like

```javascript
.catch(ErrorService.MyFirstError, function(err){ 
   // handle error
});
```

However, it appears that using a service to return constructors removes the prototype, and returns the basic object instead.

For now, we have put our error service in a separate file and are using `require()`.

Is this is intentional by sails to disallow constructor functions in Services?

chrisinajar:
As per #2791, you cannot store classes in services since it automatically binds all the methods in the service to the service's context. This breaks the class system, obviously.

Since there's no way to programatically detect the intention of a function, either class or real function, we can't fix this is a seamless automatic way. Instead I added a new flag, "noBind", to use in those weird hyper specific corner cases such as this.

Test cases included.

### DIFF
--- a/lib/hooks/moduleloader/index.js
+++ b/lib/hooks/moduleloader/index.js
@@ -555,7 +555,9 @@ module.exports = function(sails) {
         // Add a reference to the Sails app that loaded the module
         module.sails = sails;
         // Bind all methods to the module context
-        _.bindAll(module);
+        if (!module.noBind) {
+          _.bindAll(module);
+        }
       });
       return cb(null, modules);
     };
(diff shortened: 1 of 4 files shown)


tjwebb:
@chrisinajar Thanks for the PR!

Can you include a complete example and some docs on how to use this feature? It's unclear how we would explain this in the documentation on sailsjs.org. Thanks

CWyrtzen:
@chrisinajar have you had a chance to work on an example/docs on this?

PR #2911, opened 2015-05-07
Update ROADMAP.md

cburatto:
Suggestion to add `afterRead` so you can create functions to alter objects before serving them -- more specifically, stripping passwords from user objects.


tjwebb:
@cburatto you already can do this with `toJSON`: http://sailsjs.org/#!/documentation/reference/waterline/records/toJSON.html

cburatto:
Awesome, thanks for point that out! should I remove this request?

sgress454:
@cburatto Well done adding your feature request as a PR to the roadmap, definitely the right way to do it!  In this case since we have a similar feature implemented already, we'll just close this one.

PR #2920, opened 2015-05-11
Update  dependency and fix tests for iojs

listepo:


listepo:
What could be the problem? https://travis-ci.org/balderdashy/sails/jobs/63095591#L670

sgress454:
@listepo somehow `MKCALENDAR` sneaks into the response.  The version of [methods](https://github.com/jshttp/methods) that's getting installed (1.1.1) doesn't have it, so I don't know where it's coming from.  If you can get it to fail locally with a version of iojs, then you should be able to track down where that extra method is coming from...

sgress454:
The idea behind this test is that, in the absence of a custom handler, an OPTIONS request should get a response from the default Express handler, which returns the list of available methods.  This does open up the possibility of the test failing when a new version of `methods` comes out, so push comes to shove, we can just make it a less restrictive test (by just looking at a subset of the list of returned methods).  But I'd like to see what's going on here first, since this "MKCALENDAR" method is not available in any version of `methods` that I can find.

sgress454:
Dependencies are being locked for v0.12.  See https://github.com/balderdashy/sails/pull/3180#issuecomment-167150161

PR #2939, opened 2015-05-16
Adds '!' to doc links as per 9aa5083b4df55002b84b3fa8b998284b46c99f5f

justin-calleja:
Note, there are still links with this issue (e.g. in config/models.js comments). I could not find these with `egrep -R -n ".*#\/.*" .` which is what I used to find the fixed files in this commit.

Maybe someone who is familiar with Sails's src can fix those.


sgress454:
Awesome, thanks Justin.

justin-calleja:
You're welcome :smile:

PR #2944, opened 2015-05-17
Speed Up augmentModels

scott-wyatt:
Add in lodash’s quicker _.each function.

### DIFF
--- a/lib/hooks/pubsub/index.js
+++ b/lib/hooks/pubsub/index.js
@@ -71,7 +71,7 @@ module.exports = function(sails) {
 
     augmentModels: function() {
       // Augment models with room/socket logic (& bind context)
-      for (var identity in sails.models) {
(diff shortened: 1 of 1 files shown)


woollybogger:
[This jsPerf](http://jsperf.com/augmentmodels-perf-test) indicates that `_.each` is actually slightly slower than a native `for..in` loop. I'm getting an even larger difference when running that same benchmark in Node:
```
_.each x 100,933 ops/sec ±3.47% (80 runs sampled)
for in x 142,130 ops/sec ±4.07% (83 runs sampled)
```

sgress454:
I had might doubts about performance as well, but there's a better reason to implement this, which is that `_.each` is just safer than `for...in`.  This is code that runs infrequently anyway.  I'll merge pending testing.

PR #2951, opened 2015-05-20
Using the upstream consolidate.js, ensure  modules are resolved relative to the user's app's path.

aclave1:
@sgress454 @tjwebb This is the code we discussed in #2866. I've tested it with react and ejs and it appears to work the same way as before.


sgress454:
Great work Alex, thanks!

PR #3007, opened 2015-06-09
Added viewFolder option in res.locals.

zand3rs:
This is useful when you want to have multiple views for a controller and you want to dynamically switch between them just by specifying a viewFolder.
Typical usage would be to render different views for different devices especially if the device doesn't support RWD.

### DIFF
--- a/lib/hooks/views/res.view.js
+++ b/lib/hooks/views/res.view.js
@@ -182,6 +182,14 @@ module.exports = function _addResViewMethod(req, res, next) {
       layout = res.locals.layout;
     }
 
+    // Add viewFolder if provided
+    if (typeof res.locals.viewFolder !== 'undefined') {
+      relPathToView = path.join(res.locals.viewFolder, relPathToView);
+      if (typeof layout === 'string') {
+        layout = path.join(res.locals.viewFolder, layout);
+      }
+    }
+
     var pathToViews = sails.config.paths.views;
     var absPathToView = path.join(pathToViews, relPathToView);
     var absPathToLayout;


zand3rs:
May I know the status of this feature request? I've been doing a hack just to achieve this behavior and I'm hoping this pull request will solve my current problem.

If there's an existing feature that would enable me to achieve the same result I would appreciate if someone could point me to the right direction.

Thanks!

sgress454:
Are you aware that you can use the first argument of `res.view` to point to any view you like?
```
res.view("some/folder/someView", {foo: "bar"})
```
See the docs for [res.view](http://sailsjs.org/#!/documentation/reference/res/res.view.html?q=usage).

zand3rs:
Yes I'm very well aware of that. But this new option is different. Let me give you an example of my current use case to better explain.

Let's say I have a *PhotoController* with *index* and *show* actions. Using the default *res.view* behaviour for rendering I will have the following file structure:

```javascript
api/controllers/PhotoController.js
views/layout.ejs
views/photo/index.ejs
views/photo/show.ejs
```

If I want to provide a different set of views (in my case for feature phone device), I could easily do it using the viewFolder option without changing anything from my controller just by having a middleware or policy that contains a line of code like this:

```javascript
if (mobile) {
  res.locals.viewFolder = "mobile";
}
```

And then all I need to do is add the following new set of files:

```javascript
views/mobile/layout.ejs
views/mobile/photo/index.ejs
views/mobile/photo/show.ejs
```

So just imagine if I had to use the syntax you mentioned all throughout my application. Whereas I could easily get around it by using the viewFolder option and thus provide additional level of abstraction between the controllers and the views.

sgress454:
Not going to merge this for v0.12, but I'd be willing to take another look if:

1. It used `req.options.viewFolder` instead of `res.locals.viewFolder`, which would break any apps that happened to serve views with a `viewFolder` template var, and
2. It had some tests.

zand3rs:
Will make necessary changes. Thanks!

PR #3021, opened 2015-06-19
Get npm version fix

lucasconstantino:
Hello,

I was about to install Sails for contribution, and when I tried to run `npm link` I got a "Unrecognized npm version" error. This happened on the preinstall script, and was due to my npm being a prerelease version (1.5.0-alpha-4, which by the way I should update). The npm engine requirement, though, is `>= 1.4.0` which should have resolved correctly... I guess...

Anyway, while fixing this issue I ended up doing a major rewrite of the preinstall script, which did not followed most syntax pattern of other scripts in the project.

### DIFF
--- a/lib/preinstall_npmcheck.js
+++ b/lib/preinstall_npmcheck.js
@@ -1,34 +1,39 @@
-/* Npm Preinstall Script
-* This Script checks if the installed npm-Version matches the 'engine'-Item in package.json
+/**
+ * npm pre-install script.
+ *
+ * This script checks if the installed npm-version matches the required engine
+ * defined in the package.json.
  */
 
-var child_process = require('child_process'),
-    required_npm_version = require('./../package.json').engines.npm;
+
+/**
+ * [exec description]
+ * @type {[type]}
+ */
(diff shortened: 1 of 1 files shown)


tjwebb:
Why aren't we just using semver here, anyhow? https://github.com/npm/node-semver

lucasconstantino:
I had the same thought and actually went on and used node-semver the first time... but well, this script is executed on a pre-install, which means no dependency would be available yet!

sgress454:
Thanks @lucasconstantino.  In general we'd prefer not have such a large rewrite in a single commit, since it requires going through the whole thing to figure out how everything is working instead of focusing on fixing what was broken.  But seeing as this whole script is meant to perform one very specific task, and it appears to perform that task while fixing the original problem you encountered, I'm all thumbs up here.

tjwebb:
@lucasconstantino ah. chicken-egg. missed that part :)

I wonder if we can make use of `bundledDependencies` to include some core modules like semver that might help with this. (and also make download/installation faster)

PR #3023, opened 2015-06-19
support npm private modules as hooks

### LINKED ISSUE #3022: hook moduleloader does not work with private npm modules
https://github.com/balderdashy/sails/blob/master/lib/hooks/moduleloader/index.js#L444

private npm modules are installed one level below `node_modules`. e.g. for the user "balderdash"`, a module is installed in `node_modules/@balderdash/<modulename>`. The sails hook module loader does not seem to pick these up.

Changing "depth" to 3 doesn't seem to magically fix anything. I'm looking into it now.

tjwebb:
- resolves #3022

### DIFF
--- a/lib/hooks/moduleloader/index.js
+++ b/lib/hooks/moduleloader/index.js
@@ -576,9 +576,6 @@ module.exports = function(sails) {
     var name = keys[0];
     var privateModule = module[name];
 
-    console.log(module)
-    console.log(name);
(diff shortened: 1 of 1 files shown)


tjwebb:
@mikermcneil @sgress454 can you guys take a look at this and publish to npm? This is blocking on one of our projects. Thanks.

PR #3082, opened 2015-07-16
Adding whitelist filter to ActionUtil

krishnaglick:
Currently ActionUtil only has support to remove bad values based on a blacklist. While effective in most cases, I feel the addition of a whitelist can be beneficial.

This was spurred from a need in my current personal project where I was unable to directly edit the param list used by Sails (I'm mixing in google data in my controller and need the query params).

I am opening this pull request to see what the community thinks of it.

I'd like to add tests but I am unsure where they belong for ActionUtil.


krishnaglick:
So, it looks like _.omit removes everything if it's an empty array. I should have realized that last night. Will be fixing tonight when I get home from work.

mikermcneil:
Hi @krishnaglick, just getting caught up -- sorry for the slow reply.  I think the same principle from https://github.com/balderdashy/sails/pull/3102#issuecomment-172287546 applies here.  If this is still a relevant use case for you, would you mind providing some more information?

krishnaglick:
I coded around the issue by using lodash in my controller. No biggie. I was receiving some params that weren't used my by system, and were actually intended for google's api, and wanted an easy way to define what I would take.

sgress454:
Haven't seen any more chatter about this, so I'm going to close.  It's not a bad proposal by any means, but without seeing a lot more requests for the feature, and seeing as it can be worked around in userland code, it doesn't justify the maintenance burden to put it in core right now.

PR #3098, opened 2015-07-21
Have option to only populate entity primary keys.

BrighTide:
For instance, if A has 1->M with B, where A is ignorant of the relationship (i.e, A has no B field, and B has an A field) and I do a find for A, it would be awesome to have the option to be able to get a.b's as an array of ids. I would be happy to implement myself if someone can point me in the right direction :)


mikermcneil:
Hey @BrighTide, is this in the context of blueprints?

BrighTide:
Yeah, it was. We wanted to get collections as arrays of id's so we could do
a front end join and after data load and have the entities linked by
reference.

The whole project was scrapped though, and I don't know how universally
useful the feature is...

mikermcneil:
@BrighTide Gotcha.  There are some steps taken towards this using [blueprint config](http://sailsjs.org/documentation/reference/configuration/sails-config-blueprints).   If this comes up again for you, or anyone else is interested in a sort of "shallow populate" for collection associations (returning a mongo-style array of foreign keys), then please chime in here.  I think that's a reasonable proposal for the blueprints hook-- we would just need someone to [spec it out](https://github.com/balderdashy/sails/blob/master/CONTRIBUTING.md#v-proposing-features-and-enhancements) a bit further and then I think it would be actionable.  Thanks!

BrighTide:
Hey, so this has come up for me again using waterline (standalone) a company or two down the road. Is this still something you consider to be a reasonable feature? If so I have some overtime I can block out for implementation in the next 1/2 weeks.

BrighTide:
@mikermcneil I'm going to have some time this week to bust out this issue

Issue #2496, opened 2014-12-17
property policies


RWOverdijk:
Hey,

I've made a custom implementation for this, but I thought I'd run it by you guys to see if perhaps there's interest to implement it in sails.

## The concept
The concept is pretty simple. Just as with regular policies, you have policies for properties. These policies are parsed in a rule-set to make sure there's no performance loss. The rules are defined in `config/propertyPolicies.js`.

### Usage
In it's most basic form, the config could look something like this:

**config/propertyPolicies.js:**
```js
module.exports.propertyPolicies = {
  user: ['email', 'mailable'] // Whitelist for model `user`
};
```

This specifies that the _update_ blueprint is allowed to only update these two properties. Sending along id, password or any of the other properties will result in those values simply being ignored.

To achieve this, I've created a policy that does these checks for you, and populates a blacklist. Like so:

**api/policies/filterProperties.js:**
```js
var config = sails.config.propertyPolicies;

/**
 * Apply policies to models as to manage access to properties.
 *
 * @todo Allow for property filtering in find
 * @todo Allow short-hand (model: ['whitelist'])
 * @todo Allow blacklisting ({blacklist: ['blacklist']}). Use Object.getOwnProperties(sails.models[model].definition); to extract properties.
 * @todo Allow Default (*) On global, and per-model basis like with regular policies
 * @todo Allow applying policies (or even specific dir with modelPolicies?)
 */
module.exports = function(req, res, next) {
  var model = [req.options.model],
      blacklist = [];

  if (!config[model]) {
    return next();
  }

  Object.getOwnPropertyNames(sails.models[model].definition).forEach(function (property) {
    if (config[model].indexOf(property) > -1) {
      return;
    }

    blacklist.push(property);
  });

  if (!req.options.values) {
    req.options.values = {};
  }

  req.options.values.blacklist = blacklist;

  next();
};
```

Then, in `config/policies.js` just add `'filterProperties'`.

### Filtering properties
Currently this concept only applies to update. However, in the same train of thought I figured that setting up policies like this for _read_ should be possible to. This could be achieved by setting up the same config structure as policies.js currently allows, except one level deeper:

**config/propertyPolicies.js:**
```js
module.exports.propertyPolicies = {
  // Default policy for all models and properties.
  '*': false,
  User: {
    create: {
      '*'  : true,
      roles: 'isAdmin'
    },
    update: {
      mailable: 'ownsRecord',
      email   : 'ownsRecord',
[...]

PR #3102, opened 2015-07-22
RFC / FR allow different model representations

mphasize:
I'm looking for a solution to selectively show/hide or transform model properties depending on the request. Similar to removing `password` before sending a user record, but dependent on whether or not the user is authorized and who that user is.

Imagine a website that is available to public visitors, but shows more information for logged in users. Now imagine the same behavior on a RESTful JSON API. When implementing such an API the blueprints for Sails are extremely helpful, so I would like to find a solution without the need to implement custom controller actions.

One possible way of expressing this could be to support different flavors of `ModelRecord.toJSON`, like `toJSONPublic`.


RWOverdijk:
@mphasize I already made a proposal for this some time ago. You can see it here: https://github.com/balderdashy/sails/issues/2496

mphasize:
@RWOverdijk Ah, I had seen that issue some time ago but didn't remember it. Thanks for pointing it out again and yes, you certainly took the idea / proposed syntax a few steps further.

Did you ever implement it?

What I can't seem to figure out is where to bring this into a project. Policies are usually called before the blueprints and the blueprints send the data directly to responses. The model and record lifecycle hooks are all lacking enough context to do anything request specific.

I'm currently trying my hands at hacking together a custom response that takes the blueprint output and filters it. I already know it's a bad solution because it creates and extra (and in my opinion) unnecessary rev on the data, but again – I don't know anywhere else to put this.

RWOverdijk:
@mphasize I implemented it inside of the blueprint for the time being. All blueprints call this module to decide what to allow. I also have a default policy which does the checks for update, create and delete there.

Unfortunately, I wrote this code at my previous job and I can't share it. But I can write it again and make it into a hook, with some help :) My only concern was (is) performance. Applying policies per-property can be intense so this should only be used when really needed. Also the checks could be optimised a bit more.

mphasize:
@RWOverdijk Thanks for the update. In the meantime I implemented my hijack-the-response hack for this, so I have a first solution. Iterating over all the properties one at a time does sound expensive, so I would probably look for a different way. However, good to know that you also choose the blueprint route for implementing this.

For reference, here's what I did:
```javascript
// api/responses/ok.js
/**
 * 200 (OK) Response
 *
 * Usage:
 * return res.ok();
 * return res.ok(data);
 * return res.ok(data, 'auth/login');
 *
 * @param  {Object} data
 * @param  {String|Object} options
 *          - pass string to render specified view
 */

var actionUtil = require( 'sails/lib/hooks/blueprints/actionUtil' );

function capitaliseFirstLetter( string ) {
	return string.charAt( 0 ).toUpperCase() + string.slice( 1 );
}

function sendOk( req, res, data, options ) {
	// Get access to `req`, `res`, & `sails`
	var sails = req._sails;
	sails.log.silly( 'res.ok() :: Sending 200 ("OK") response' );

	// Set status code
	res.status( 200 );

	// If appropriate, serve data as JSON(P)
	if ( req.wantsJSON ) {
		return res.jsonx( data );
	}

	// If second argument is a string, we take that to mean it refers to a view.
	// If it was omitted, use an empty object (`{}`)
	options = ( typeof options === 'string' ) ? {
		view: options
	} : options || {};

	// If a view was provided in options, serve it.
	// Otherwise try to guess an appropriate view, or if that doesn't
	// work, just send JSON.
	if ( options.view ) {
		return res.view( options.view, {
			data: data
		} );
	}

	// If no second argument provided, try to serve the implied view,
	// but fall back to sending JSON(P) if no view can be inferred.
	else return res.guessView( {
		data: data
	}, function couldNotGuessView() {
		return res.jsonx( data );
	} );
}

module.exports = function transformResponse( data, options ) {

	// Get access to `req`, `res`, & `sails`
	var req = this.req;
	var res = this.res;
	var sails = req._sails;

	var Model = actionUtil.parseModel( req );
	var blueprint = capitaliseFirstLetter( req.options.action ); // we actually enable this for all actions, not only blueprints

	if ( typeof Model[ "controlAfter" + blueprint ] === "function" ) {
		return Model[ "controlAfter" + blueprint ]( req, res, data, options, sendOk );
	} else {
		sails.log.debug( 'OK afterBlueprint: Model "' + Model.identity + '" has no method controlAfter' + blueprint + '.' );
		return sendOk( req, res, data, options );
	}

};
 ```

This hacked response enables usage of `controlAfterFind`, `controlAfterFindOne` and `controlAfterPopulate` hooks on the model with a `fn(req,res,data,options,callback)` signature.
This matches nicely with my `beforeBlueprint` [policy](https://gist.github.com/mphasize/e9ed62f9d139d2152445).

Still, only a workaround...

mikermcneil:
@mphasize @RWOverdijk I think the biggest challenge/risk with this sort of a feature is deciding (for a given endpoint) when it's time to move away from decorating a blueprint action and to just defining a custom action.  Sometimes, you can go all the way to production for a given endpoint using only blueprints + a few minor customizations.  My concern recently has been, if we provide too much customizability for blueprints, we create the possibility for a scenario where you get stuck between a rock and a hard place, e.g.:

> "uurggh I've put in so much effort making this endpoint work with blueprints!  I can't stop now!" (when it would actually be much more efficient to just write a custom action)

That said, I think this is an important topic to explore further.  In my experience, I've _always_ found it better to disable all blueprint routing in production, _but_ there are certain use cases where blueprint **actions** are useful up to your app's first production launch and beyond.

It would be great if both of you guys could provide a little background of how you are using blueprints vs. custom actions today (i.e. what kind of app, what kind of endpoint).  Obviously, no need to share specifics if it's something that needs to be kept confidential for work, but if you wouldn't mind, knowing more about the backdrop here would be helpful.

PR #3108, opened 2015-07-24
Upgraded EJS to v2 from v0

uzyn:
For #3066


jetsonjohn:
I am facing some issue on upgrading see #3242

PR #3115, opened 2015-07-27
Allow to set using SSL in http config

JumpLink:
Allow to set using SSL in sails.config.http.serverOptions

### DIFF
--- a/lib/hooks/http/initialize.js
+++ b/lib/hooks/http/initialize.js
@@ -33,12 +33,13 @@ module.exports = function(sails) {
         // Create express server
         var app = sails.hooks.http.app = express();
         app.disable('x-powered-by');
-        // (required by Express 3.x)
-        var usingSSL = sails.config.ssl.key && sails.config.ssl.cert;
 
         // Merge SSL into server options
         var serverOptions = sails.config.http.serverOptions || {};
         _.extend(serverOptions, sails.config.ssl);
+        
+        // (required by Express 3.x)
+        var usingSSL = serverOptions.usingSSL || (sails.config.ssl.key && sails.config.ssl.cert);
 
         // Get the appropriate server creation method for the protocol
         var createServer = usingSSL ?


JumpLink:
I need this because I use my `serverOptions` tu support [SNI](https://stackoverflow.com/questions/12219639/is-it-possible-to-dynamically-return-an-ssl-certificate-in-nodejs) with the SNICallback and I do

    var serverOptions = {
      SNICallback: function (domain, cb) {
        console.log("SNICallback from domain:", domain);
        MultisiteService.getCurrentSiteConfig(domain, function (err, siteConf) {
          if(err) return cb(err);
          var dirname = MultisiteService.getSitePathFromSiteConf(siteConf);
          var ca_path = path.join(dirname, siteConf.ssl.ca);
          var key_path = path.join(dirname, siteConf.ssl.key);
          var cert_path = path.join(dirname, siteConf.ssl.cert);

          sails.log.debug(dirname, ca_path ,key_path, cert_path);

          var ctx = tls.createSecureContext({
            ca: fs.readFileSync(ca_path),
            key: fs.readFileSync(key_path),
            cert: fs.readFileSync(cert_path),
            passphrase: siteConf.ssl.passphrase
          });

          cb(null, ctx);

        });
      }
    };

And I do not neet to set `key` and `cert` in my local.js, but I can work it around with just set them to true.

sgress454:
@JumpLink just to be clear, are you saying the current workaround for your issue is to set `sails.config.ssl.key` and `sails.config.ssl.cert` to `true`?  Because I'd almost rather just document that than add a new `usingSSL` config that would be a little confusing to document...

JumpLink:
Yes this is true, take a look at my [config/local.js](https://github.com/JumpLink/cms/blob/master/config/local.js) as an example.

This is necessary, because sails checks if this variables are set:
```var usingSSL = sails.config.ssl.key && sails.config.ssl.cert;```

This is the only way I can use my custom 'serverOptions' with ssl, see [config/http.js](https://github.com/JumpLink/cms/blob/master/config/http.js)
If you do not need your own server options, this is not necessary. 

My patch would fix this problem as you can activate ssl with ```serverOptions.usingSSL = true;```.

sgress454:
@JumpLink, ended up simplifying this a bit by allowing you to just set `sails.config.ssl` to `true`: https://github.com/balderdashy/sails/commit/23e4d4df1293e00d63b2c46f1a1b87d77631c4e1#diff-b6a9d24b106300f5b5fda5b3eb4d093aR34

Also added some tests for https.  Thanks for the PR though!

PR #3123, opened 2015-07-29
DOCS: Added hooks to MODULES

listepo:


sgress454:
Thanks for the PR.  There are currently [109 hooks and counting](https://www.npmjs.com/search?q=sails-hook) for Sails, so putting them all on the MODULES page would be a little overwhelming...maybe just a link to that NPM page instead?

listepo:
I agree

listepo:
When it created an alternative to http://www.emberaddons.com or http://emberobserver.com?

PR #3126, opened 2015-07-30
update blueprint: prevent updating PK, but leave it for Model lifecycle callbacks

jakutis:
might crash your connection provider. tested only on sails-mongo.


sgress454:
Can you put the warning about not changing the primary key back in?

jakutis:
Yes, again misunderstood which warning was wanted. Restored.

sgress454:
I meant the console warning--it's cool, I took care of it.  Thanks for your help!

jakutis:
oooh.. ok.

Thank you for making Sails!

PR #3128, opened 2015-07-30
ORM hook: teardown all connections

jakutis:
sails-mongo adapter, for example, does not actually teardown the db
connection, if conn is null:
https://github.com/balderdashy/sails-mongo/blob/30034f3d927920d9c5ac6d9747fe28581cbda1b1/lib/adapter.js#L133


tjwebb:
@jakutis it's the adapter's job to teardown properly. If it isn't, I would file an issue in sails-mongo.

jakutis:
@tjwebb right, created PR on sails-mongo: https://github.com/balderdashy/sails-mongo/pull/315

PR #3141, opened 2015-08-07
Out of the box support for babel/ES6?

internalfx:
Babel supports `[".es6", ".es", ".jsx", ".js"]` file extensions.

Can't include `js` in that list or it will always try to require. But the others should be fine.

Thoughts?


internalfx:
Don't merge this.

PR #3166, opened 2015-08-20
Views hook refactor

### LINKED ISSUE #3041: sails still tries to load ejs module even when views hook is disabled
If I disable the "views" hook, I still get this error when lifting the app:

```sh
Could not find module: ejs
```

It's benign, but looks scary. Trying to set config.views.engine to null causes worse problems.

tjwebb:
- closes #3041 
- `async` is way overused everywhere. removed occurrence from views hook
- de-fang missing `ejs` module warning
- reformatted to match the rest of the codebase
- use `_.isString` instead of `typeof`

### DIFF
--- a/lib/hooks/views/actions.js
+++ b/lib/hooks/views/actions.js
@@ -1,10 +1,4 @@
-/**
- * Module dependencies
- */
-
 var _ = require('lodash');
-
-
 
 /**
  * Load views and generate view-serving middleware for each one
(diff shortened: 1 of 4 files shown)


sgress454:
Clearly we need a test around it that should have failed here, so that's our bad.  The async usage is not superfluous--the layout shim needs to wait until after the http hook (i.e. Express) loads before continuing, otherwise the view engine is added without it.  The upshot is that if you lift a Sails app after this commit using .ejs templates, they won't use the layout at all.  You can try it with a new app--it'll look okay in the browser, but the socket client won't load and if you check the source you'll see it's just the body of `homepage.ejs` without the `layout.ejs` wrapper (so no linked assets).

I'm not sure which part of this helps with #3041, so I don't want to just revert the whole thing, but we need to put the asynchronous usage back.  For my part I'll add a test to check that the layout is rendered with the default usage.

tjwebb:
> the layout shim needs to wait until after the http hook (i.e. Express) loads before continuing, otherwise the view engine is added without it.

Why can't it just listen on `sails.after('hook:http:listening`, ...)` ?

sgress454:
At some point, it was determined that in order for the layouts to work, `ejs-locals` had to be loaded and saved as `sails.config.views.engine.fn` _after_ the `http` hook loaded.  I think it was because otherwise, Express was overwriting the value when it loaded the view engine.  A very quick test made it seem like that is no longer the case, so it's possible we can just remove the  `sails.after('hook:http:loaded')` wrapper from around the `ejs-locals` loading code entirely.  But with your patch, you're _still_ waiting for the `http` hook to load before loading `ejs-locals`, but in the meantime, the `views` hook declares itself loaded (because it doesn't have to wait for the layout shim anymore, since the `async.auto` is replaced by synchronous calls), and the `http` hook gets that notification and sets the view engine as-is, in its pre-layout-shim state.  You can see for yourself that the layout doesn't load.  I added a test in https://github.com/balderdashy/sails/commit/60efc454bb3b7d61057636299ccfb97e1f6e2944 that is failing now, and should start passing once this is fixed up.

PR #3168, opened 2015-08-20
add "hookName" option to installable hook config

tjwebb:
- cleaned up formatting
- set "hookName" in order to specify installable hook name.
e.g.
```
{
  "sails": {
    "isHook": true,
    "hookName": "myHook"
  }
}
```

### DIFF
--- a/lib/hooks/moduleloader/index.js
+++ b/lib/hooks/moduleloader/index.js
@@ -1,23 +1,10 @@
 module.exports = function(sails) {
 
-
-  /**
-   * Module dependencies
-   */
-
   var path = require('path');
   var async = require('async');
   var _ = require('lodash');
   var buildDictionary = require('sails-build-dictionary');
   var walk = require('walk');
-
-
-  // TODO:
-  // Look at improving `includeAll` to work asynchronously
-  // CommonJS `require` is a blocking operation, and makes apps
-  // start slower.
-
-
 
   /**
    * Module loader
@@ -25,9 +12,6 @@ module.exports = function(sails) {
    * Load a module into memory
    */
   return {
-
-
-    // Default configuration
     defaults: function (config) {
 
       var localConfig = {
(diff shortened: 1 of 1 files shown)


sgress454:
Won't this allow two NPM-published hooks to have conflicting names?

tjwebb:
Currently, sails treats the hooks `sails-hook-foo` and `foo` identically. The `hookName` option gives the hook an opportunity to de-clutter the namespace and avoid conflicts with other hooks.

sgress454:
Gotcha--yeah the idea was to make the `sails-hook-` prefix the best practice for creating Sails hooks (that way you can easily find them on Github or npmjs.org).  What you did is fine, but it could catch someone off-guard if they install `sails-hook-email` and `some-awesome-email-hook` and the latter sets its hook name to `email` (causing a conflict).  Either way, you can change the hook name on a per-app basis using `sails.config.installedHooks['some-awesome-email-hook'].name`--probably should document that somewhere...

tjwebb:
Yea, a lot of the docs around hooks isn't as good as it could be. I'd like to work on improving this as well. 

As far as naming: I'd like to adopt the strategy of http://react-components.com, where they mine npm for a `react-component` keyword rather than relying on a naming prefix. npm now supports namespaces, so `@orgA/sails-hook-foo` would be the same as `@orgB/sails-hook-foo`. So my assertion is that naming conflicts are already possible, though as a practical matter, not really a burning issue. The `hookName` feature at least gives hook the opportunity to resolve these conflicts.

sgress454:
> probably should document that somewhere...

 I stand corrected, it's [here](http://sailsjs.org/documentation/concepts/extending-sails/hooks/using-hooks#?changing-the-way-sails-loads-an-installable-hook)

> Yea, a lot of the docs around hooks isn't as good as it could be. I'd like to work on improving this as well.

Knock yourself out! :)

PR #3183, opened 2015-08-25
Remove unused variables and modules

listepo:
We need a hard Code style and enabled jshint or replaced by eslint


package.json
```diff
@@ -58,13 +58,11 @@
     "grunt-sails-linker": "~0.10.1",
     "grunt-sync": "~0.2.3",
     "i18n": "~0.5.0",
-    "include-all": "~0.1.6",
```
tjwebb: `include-all` is definitely used. Am I misreading the diff?

listepo: This module is not used in the core, [see](https://github.com/balderdashy/sails/search?utf8=%E2%9C%93&q=include-all). But used in [sails-build-dictionary](https://github.com/balderdashy/sails-build-dictionary/blob/master/package.json#L19)

listepo:
Ping

package.json
sgress454: It's used by the sails-new generator (it's copied into new projects).  Without it, Grunt fails on new projects.

PR #3223, opened 2015-09-11
Fix blueprints puralize option

Benew:
don't call `pluralize('/quiz')` (returns `/quizs`). call `pluralize('quiz')` (returns `quizzes`).

### DIFF
--- a/lib/hooks/blueprints/index.js
+++ b/lib/hooks/blueprints/index.js
@@ -243,15 +243,16 @@ module.exports = function(sails) {
 
 
         // Determine base route
-        var baseRoute = config.prefix + '/' + controllerId;
-        // Determine base route for RESTful service
-        // Note that restPrefix will always start with /
-        var baseRestRoute = config.prefix + config.restPrefix + '/' + controllerId;
+        var baseRouteName = controllerId;
 
         if (config.pluralize) {
-          baseRoute = pluralize(baseRoute);
-          baseRestRoute = pluralize(baseRestRoute);
+          baseRouteName = pluralize(baseRouteName);
         }
+        
+        var baseRoute = config.prefix + '/' + baseRouteName;
+        // Determine base route for RESTful service
+        // Note that restPrefix will always start with /
+        var baseRestRoute = config.prefix + config.restPrefix + '/' + baseRouteName;
 
         // Build route options for blueprint
         var routeOpts = config;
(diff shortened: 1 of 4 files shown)


listepo:
+1

tjwebb:
@Benew mind including a couple unit tests?

sylvainlap:
Tests on pluralized routes already exists:

https://github.com/balderdashy/sails/blob/master/test/integration/router.APIScaffold.test.js#L186

and they pass.

Benew:
@tjwebb what should we do?

mikermcneil:
> Tests on pluralized routes already exists:
> 
> https://github.com/balderdashy/sails/blob/master/test/integration/router.APIScaffold.test.js#L186
> 
> and they pass.

Please add tests that demonstrate the breaking behavior (i.e. that fail when your patch is not in place).  Thanks!

Benew:
This is weird... The tests are failing because of an unrelated timeout on node v0.12

Benew:
Thanks @sylvainlap ! Looking forward to the merge!

Benew:
When can we hope for a merge?

Btw, would be more than happy to contribute more since some anchors are leaving the ship ;-)

tjwebb:
@Benew we're leaving to work on Trails.js: https://github.com/trailsjs. We'd love to have your help over there. No reason to try to catch a falling knife.

Benew:
@tjwebb So you think I should wait and move my project to trails? What's the roadmap?
I think a lot of us will want to keep using sails for production apps. And some of us might want to maintain sails. Don't go without leaving the keys under the mat...

tjwebb:
> I think a lot of us will want to keep using sails for production apps. And some of us might want to maintain sails. Don't go without leaving the keys under the mat...

Sails is a legacy platform. Express3 is EOL, and there's many fundamental problems in the framework. Feel free to keep using it, but I wouldn't invest too much time in improving the Sails framework itself. The next "version" of Sails will actually be called Trails.js, and there will be an upgrade path.

sgress454:
Thanks @Benew, sorry we're just getting around to this.  Rest assured that Sails is under active development and we have no plans to stop!  We're in the process of going over all of the merged PRs since the last release to make sure things are stable for v0.12.  Re: Express 3, we're using it because it's stable and tested with Sails, and we're committed to maintaining any issues that arise with it until we are able to follow through on our plan to separate the router out entirely, at which point you'll be able to use Express 4 or whatever other backing server you desire (see https://github.com/balderdashy/sails/pull/3235#issuecomment-141584785).

This PR looks good, just running the tests and I'll merge.  Thanks for your patience.

Issue #2977, opened 2015-06-01
Getting 400 when posting JSON array to blueprint route


luisincrespo:
Hi.

I'm using Sails 0.11. Your docs say it is possible to post multiple objects to a blueprint route, just passing a JSON array.

I'm trying to do this, but I'm getting 400 bad request as response.

This is my case:

I have a model called *Prediction*, which only has two fields, *user* (id) and *poll* (id). I'm posting this (to /prediction):

    [
        {
            "user": "556b33e884eb522f02a4f52e",
            "poll": "556c4ff8243f03e501e2b65c" 
        }
    ]

And I'm getting a 400 response, with the following body:

>{
  "error": "E_VALIDATION",
  "status": 400,
  "summary": "2 attributes are invalid",
  "model": "Prediction",
  "invalidAttributes": {
    "user": [
      {
        "rule": "required",
        "message": "\\\"required\\\" validation rule failed for input: null"
      }
    ],
    "poll": [
      {
        "rule": "required",
        "message": "\\\"required\\\" validation rule failed for input: null"
      }
    ]
  }
}

ArtworkAD:
:+1: the same for me. I cannot assign an empty array. It fails with
```
{
    "error": "E_VALIDATION",
    "status": 400,
    "summary": "1 attribute is invalid",
    "model": "Ticket",
    "invalidAttributes": {
        "notifications": [
            {
                "rule": "required",
                "message": "\"required\" validation rule failed for input: []"
            }
        ]
    }
}
```

mattcodez:
The blueprint [uses](https://github.com/balderdashy/sails/blob/e87aed76363357cade682f75c179da8c9c1e4954/lib/hooks/blueprints/actionUtil.js#L274) req.params.all() which parses a POSTed JSON array as an object
Example:
```
{ '0': { name: 'q1', value: '3' },
  '1': { name: 'q2', value: '5' } }
```
Whereas `req.body` parses as expected
```
[ { name: 'q1', value: '3' }, { name: 'q2', value: '5' } ]
```

Not quite sure how this should be fixed. Perhaps check to see if `req.body` is an array first and then default to `req.params.all()` if not?

Hanifb:
Same for me. Without the possiblity to create multiple objects on one post, Sails.js is worthless.

CiscoKidxx:
That's a tad extreme and just your narrow minded opinion.

Hanifb:
[...]

PR #3228, opened 2015-09-14
Added support for JSON arrays

Hanifb:
Fixes the issue balderdashy/sails#2977, which sails would convert an
array to an object and subsequently fail validation.

### DIFF
--- a/lib/hooks/blueprints/actions/create.js
+++ b/lib/hooks/blueprints/actions/create.js
@@ -40,7 +40,7 @@ module.exports = function createRecord (req, res) {
 				Model.subscribe(req, newInstance);
 				Model.introduce(newInstance);
 			}
-			Model.publishCreate(newInstance.toJSON(), !req.options.mirror && req);
+			Model.publishCreate(newInstance, !req.options.mirror && req);
 		}
 
 		// Send JSONP-friendly response if it's supported
--- a/lib/hooks/blueprints/actionUtil.js
+++ b/lib/hooks/blueprints/actionUtil.js
@@ -270,6 +270,18 @@ var actionUtil = {
       throw new Error('Invalid `req.options.values.blacklist`. Should be an array of strings (parameter names.)');
     }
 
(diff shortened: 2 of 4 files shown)


tjwebb:
nice work @Hanifb

sgress454:
@Hanifb awesome--a test for this in `/test/integration/router.APIScaffold.test.js` would allow us to merge...

Hanifb:
@sgress454 Done!

tjwebb:
@dominicrico @Hanifb can you guys come up with a test that ensures this scenario passes?

dominicrico:
@Hanifb are you doing this?

Hanifb:
Already did, se my PL on sails-hook-sockets: https://github.com/balderdashy/sails-hook-sockets/pull/10

dominicrico:
:thumbsup:

lib/hooks/blueprints/actions/create.js
```diff
@@ -40,7 +40,7 @@ module.exports = function createRecord (req, res) {
 				Model.subscribe(req, newInstance);
 				Model.introduce(newInstance);
 			}
-			Model.publishCreate(newInstance.toJSON(), !req.options.mirror && req);
+			Model.publishCreate(newInstance, !req.options.mirror && req);
 		}
```
sgress454: @Hanifb What was the thinking behind this?  Is it somehow necessary to get arrays to work?  By pushing this logic into `publishCreate`, we force users of that method to pass in model instances instead of being able to send in plain objects.  Sorry I didn't catch this earlier.

Hanifb: Wow, i dont even remember. Actually it is kinda unneccessary, because the publishCreateSingle//former publishCreate  already does that.

Hanifb: @sgress454 Well now i remember, the reason is because how pluralize method works, it needs an instance of a model. Do you think we should remove the call to pluralize and instead use a simple    `if(!_.isArray(instance)){ [instance] }`?

PR #3230, opened 2015-09-14
Added dependencies status image - take II

tarlepp:
This is a second PR of the same change. For some reason @mikermcneil just removed previously accepted PR on commit 5c3a23c6e8d4679cf28d3ed153c4fd85274f8155 with comment "removed broken badge". Actual badge isn't broken, only the dependencies that sails uses are outdated.

Imho isn't good policy to "hide" things like this...

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 [![Gitter](https://badges.gitter.im/Join Chat.svg)](https://gitter.im/balderdashy/sails?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
 [![NPM version](https://badge.fury.io/js/sails.svg)](http://badge.fury.io/js/sails)
 [![Build Status](https://api.travis-ci.org/balderdashy/sails.png?branch=master)](https://travis-ci.org/balderdashy/sails)
-
+[![Dependency Status](https://david-dm.org/balderdashy/sails.svg)](https://david-dm.org/balderdashy/sails)
 
 Sails.js is a web framework that makes it easy to build custom, enterprise-grade Node.js apps. It is designed to resemble the MVC architecture from frameworks like Ruby on Rails, but with support for the more modern, data-oriented style of web app development. It's especially good for building realtime features like chat.
 


tjwebb:
> Imho isn't good policy to "hide" things like this...

I agree. Accountability is one of the things that drives improvement. 

@mikermcneil do you have any objections to this PR?

sgress454:
I think it was broken when he removed it.

tjwebb:
This is the link: https://david-dm.org/balderdashy/sails.svg

Works just fine.

mikermcneil:
@tjwebb @sgress454 @tarlepp Hey guys, it was showing up as a broken image about 50% of the time I refreshed, so I removed it because it just looked a little trashy.  Maybe their servers have been spotty?

mikermcneil:
@tarlepp @tjwebb Also, aside from the practical reason mentioned above-- there is something else I've been meaning to bring up for a while:

I appreciate the interest in making sure our dependencies are stable. However, there's nothing inherently wrong with dependencies being locked to a particular version.  In fact, this is good-- it means that an accidental NPM publish from the maintainer of a different project can never completely wreck our apps (in fact, I'd encourage you to lock your Sails dependency version to the `^0.11.x` semver range-- you _really don't want_ updates to your dependencies that contain breaking changes being picked up automatically when you run NPM install).

The problems start, in my opinion, when outdated dependencies present an unexpected security risk or serious performance issue. With that said-- If any of the dependencies [David flags as "out of date"](https://david-dm.org/balderdashy/sails#info=devDependencies&view=table) warrant that kind of investigation, let us know and we'll take care of it ASAP.

So, based on the comments above, it doesn't really make sense to keep the badge on our README unless the "out-of-date" dependencies it's referring to are actually dependencies we need to update.  I don't mean to offend, but the core team has no intention of updating the dependencies of Sails just for the sake of updating-- and since, as far as I know, that's how the David badge calculates "out-of-date-ness", I won't put a big red icon on the readme.  It could trouble our community, or worse, could confuse someone new to Node and/or NPM.

My apologies for not being clear about this in the past.  And _please_ let me know if you encounter performance or stability issues in any of our dependencies that require an upgrade.

Issue #3176, opened 2015-08-24
sails fails with npm3 flatten deps


valery-barysok:
due to flatten dependencies sails fails to lift application. It can be easily fixed by replacing `express/node_modules/connect` with just `connect` in two places. but I do not know if it is right solution.

```js
[32minfo: [39m[90mStarting app...[39m

module.js:338
    throw err;
          ^
Error: Cannot find module 'express/node_modules/connect'
    at Function.Module._resolveFilename (module.js:336:15)
    at Function.Module._load (module.js:278:25)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
    at Object.<anonymous> (...\node_modules\sails\lib\hooks\session\index.js:25:22)
    at Module._compile (module.js:460:26)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
    at ...\node_modules\sails\lib\app\configuration\index.js:63:34
    at ...\node_modules\sails\node_modules\lodash\dist\lodash.js:3741:15
    at forOwn (...\node_modules\sails\node_modules\lodash\dist\lodash.js:2106:15)
    at Function.reduce (...\node_modules\sails\node_modules\lodash\dist\lodash.js:3738:9)
    at Configuration.defaultConfig (...\node_modules\sails\lib\app\configuration\index.js:58:18)
```

amyspark:
I got the same error, as a workaround please go to the `express` folder in `node_modules` and run `node install` there.

Alaneor:
Just FYI - npm 3 is now marked as *latest*, meaning it is out of beta. People who use Sails.js and upgrade npm **will break** their deps. Perhaps a patch release for 0.11 would save the day?

Also see the npm 3 announcement [here](http://blog.npmjs.org/post/129378362260/npm-weekly-29-npm-3-out-of-beta-nick-out-of-the).

I tried to disable the hooks which use the offending statement (session, csrf), but it seems they are loaded even though they are not used.

tjwebb:
> People who use Sails.js and upgrade npm will break their deps

Is this issue caused merely by using npm 3?

Alaneor:
They need to install their dependencies using npm 3. That would cause the deps to be flattened and the modules mentioned previously will simply not be at the specified path.

adambiggs:
[...]

PR #3245, opened 2015-09-19
Remove nested require in csrf to accomodate npmv3

calling:
Fixes https://github.com/balderdashy/sails/issues/3176

### DIFF
--- a/lib/hooks/csrf/index.js
+++ b/lib/hooks/csrf/index.js
@@ -77,8 +77,14 @@ module.exports = function(sails) {
           var allowCrossOriginCSRF = sails.config.csrf.origin.split(',').map(trim).indexOf(req.headers.origin) > -1;
 
           if (sails.config.csrf.protectionEnabled) {
-            var connect = require('connect');
-
+            var connect;
+            try {
+              // Support npm v3 flat dependency installations.
+              connect = require('connect');
+            } catch (err) {
+              // Contiue to support npm v2 dependency installations.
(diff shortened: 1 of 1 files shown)


lib/hooks/csrf/index.js
```diff
@@ -77,8 +77,14 @@ module.exports = function(sails) {
           var allowCrossOriginCSRF = sails.config.csrf.origin.split(',').map(trim).indexOf(req.headers.origin) > -1;
 
           if (sails.config.csrf.protectionEnabled) {
-            var connect = require('express/node_modules/connect');
-
+            var connect;
+            try {
+              // Support npm v3 flat dependency installations.
+              connect = require('connect');
```
tjwebb: 1) Isn't this the express 4 connect? Does it work with express 3?
2) I think we'd also want to add this to `optionalDependencies` at least, otherwise things get confusing.

calling: This has more to do with the way npm 3 installs dependencies. With npm 3, a fresh clone of sails (at the current version of express of 3.x), `npm install` will install express' connect dependency adjacent to the express directory. See npm's [3.0.0 release notes](https://github.com/npm/npm/blob/master/CHANGELOG.md#flat-flat-flat) for more details.

tjwebb: Alright I'll check that out, thanks.

sgress454:
Thanks for the PR @calling.  We ended up just making `connect` a direct dependency of Sails!

Issue #2995, opened 2015-06-05
Populating despite populate:false / performance implications of associations and PubSub 


leobudima:
I have a Courier model, with one-to-many association to Deliveries model:

```
Courier.js

...
deliveries: {
	collection: 'delivery',
	via: 'courier'
},
...

```

Even though I have set `populate: false` in blueprints.js, and never make any `populate()` calls, every 15 seconds I update the Courier model (his location, in this case), Sails sends back a fully populated Courier record with 100s of associated Delivery records (which first need to be fetched from the database!)...

The "culprit" is `hooks/blueprints/actions/update.js`, specifically `updateOneRecord`, which contains:

```
Model.findOne(pk).populateAll().exec(function found(err, matchingRecord) {
```
While I understand the convenience of doing things like this and that some people definitely expect it, (and I realise the possibility of overriding update method in controller) - if populate is set to false, wouldn't it make sense not to automatically populate in this case either?

Thanks!

tjwebb:
> Sails sends back a fully populated Courier record with 100s of associated Delivery records

The `defaultLimit` is set to `30`.

>  if populate is set to false, wouldn't it make sense not to automatically populate in this case either?

Yea, I agree with this. There should be a conditional statement which only does populateAll if you have `populate: true`. Since I think you've accurately diagnosed the problem, would you mind submitting a PR?

leobudima:
Thanks a lot for the comment @tjwebb - you're right, I completely missed the `defaultLimit` that was previously set rather high in our case :) Definitely makes it much less of a problem by default.

I'll happily submit a PR shortly!

sailsbot:
Thanks for posting, @leobudima.  I'm a repo bot-- nice to meet you!

It has been 60 days since there have been any updates or new comments on this page.  If this issue has been resolved, feel free to disregard the rest of this message.  On the other hand, if you are still waiting on a patch, please:

  + review our [contribution guide](https://github.com/balderdashy/sails/blob/master/CONTRIBUTING.md) to make sure this submission meets our criteria (only _verified bugs_ with documented features, please;  no questions, commentary, or bug reports about undocumented features or unofficial plugins)
  + create a new issue with the latest information, including updated version details with error messages, failing tests, and a link back to [the original issue]().  This allows GitHub to automatically create a back-reference for future visitors arriving from search engines.

Thanks so much for your help!

PR #3248, opened 2015-09-22
"Update" blueprint honors model's "populate" configuration

leobudima:
Here's a quick fix for: https://github.com/balderdashy/sails/issues/2995, which doesn't return populated models on update if populate was set to false.


sgress454:
Thanks @leobudima.  Updated a bit to use `populateRequest` and added some missing tests; merging now!

leobudima:
Thanks, great!

PR #3300, opened 2015-10-11
added support for cassandra as session store

angleto:
this code enable the possibility to use cassandra as session store, a pull request with templates modifications is going to be open on sails-generate-backend repository.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 * [ENHANCEMENT] Support CSRF hook route configuration [#2366](https://github.com/balderdashy/sails/issues/2366)
 * [BUGFIX] Fix [RangeError: Maximum call stack size exceeded] error in PubSub hook
 * [ENHANCEMENT] Support layout for Ractive template engine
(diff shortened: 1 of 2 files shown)


tjwebb:
Awesome! @angleto can you include some tests?

angleto:
@tjwebb sure I could add some test but I need a cassandra service running, I can set up an instance of  cassandra using travis, it is ok? Also, I do not see the service mongo and the tests for mongo session store in this repository, this test is just missing or it is on a different repo?

mikermcneil:
@angleto that's right, we also need tests for the mongo session store. If you wouldn't mind setting those up on Travis CI, that'd be very helpful.

mikermcneil:
@angleto Came back around to take a closer look at this, and it looks like you should be able to use https://github.com/B2MSolutions/connect-cassandra (a connect session adapter).  I'm working on docs for configuring custom session adapters in Sails today, so please let me know (open a new issue in [sails-docs](github.com/balderdashy/sails-docs)) if there are any special considerations to keep in mind here.  Thanks!

PR #3322, opened 2015-10-22
Update hook.cors_csrf unit test to be compliant with node 4.x.x

AlexisNo:
The unit test suite didn't pass with node 4.x.x. [The MKCALENDAR http method is now supported](https://github.com/nodejs/http-parser/issues/255) but the test was checking the response of an OPTION request globally.

In this PR, I checked the presence of each http method that should be available for node < 4.x.x.
MKCALENDAR is not tested, but the test suite works successfully with 4.x.x .


sgress454:
Thanks for the PR @AlexisNo.  We ended up opting for an approach that mimics the underlying `methods` module to return the full list of methods, since the objective of the test is to determine as much as possible that the default Express handler is being used.

PR #3330, opened 2015-10-25
Added dependencies status images

tarlepp:


sgress454:
Please see https://github.com/balderdashy/sails/pull/3180#issuecomment-167150161.  Dependencies are locked for the next release based on actual testing with Sails and inspection of their changelogs (if available).

PR #3336, opened 2015-10-27
Preserve context when calling log functions

Alaneor:
This fixes a rare issue where the built-in `sails.log` object is overridden with a custom logging implementation which uses functions that depend on a particular value of `this`.

For example, if a `sails.log.verbose()` function is called like this:

```js
var log = sails.log.verbose

log(‘test message’)
```

The value of `this` within this function will be either `undefined` (strict mode) or the `global` object, instead of the expected `sails.log` object.

> For the sake of interest, I have replaced the logger with a [bunyan](https://github.com/trentm/node-bunyan) logger, which is implemented this way.

### DIFF
--- a/lib/app/lower.js
+++ b/lib/app/lower.js
@@ -58,7 +58,6 @@ module.exports = function lower(cb) {
     // Shut down Socket server
     // wait for all attached servers to stop
     sails.emit('lower');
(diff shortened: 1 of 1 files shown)


tjwebb:
sorry, didn't mean to close this.

Issue #3314, opened 2015-10-20
In sails 0.11.2 exposeHeaders is missing


smanne:
exposeHeaders code is missing in cors in v0.11.2 it exists in v0.11.1. Please solve this.

smanne:
Please find cors/index.js in tag v0.11.2 
https://github.com/balderdashy/sails/blob/v0.11.2/lib/hooks/cors/index.js

Same file in master branch
https://github.com/balderdashy/sails/blob/master/lib/hooks/cors/index.js (L:184)

britton-jb:
@tjwebb I've got a fix for this, including the test, but it doesn't look like I can create a PR against v0.11.2. Should I just create the PR against master instead to add the test?

tjwebb:
yea! Go ahead and submit to master, that'd be awesome.

sgress454:
@britton-jb You can submit a PR against the `stable` branch if you have a fix for 0.11.2!

sailsbot:
Thanks for posting, @smanne.  I'm a repo bot-- nice to meet you!

It has been 30 days since there have been any updates or new comments on this page.  If this issue has been resolved, feel free to disregard the rest of this message.  On the other hand, if you are still waiting on a patch, please:

  + review our [contribution guide](https://github.com/balderdashy/sails/blob/master/CONTRIBUTING.md) to make sure this submission meets our criteria (only _verified bugs_ with documented features, please;  no questions, commentary, or bug reports about undocumented features or unofficial plugins)
  + create a new issue with the latest information, including updated version details with error messages, failing tests, and a link back to [the original issue]().  This allows GitHub to automatically create a back-reference for future visitors arriving from search engines.

Thanks so much for your help!

PR #3337, opened 2015-10-27
Adds test for exposeHeader cors config

britton-jb:
In response to issue #3314, adds a test for the exposeHeaders CORS config.


sgress454:
@britton-jb Thanks for your submission.  The code and the test for `exposeHeaders` is already there in the master branch.  I understand it doesn't exist in 0.11.2; you can submit a PR against the stable branch to fix that.  I think the issue is that 0.11.1 was never intended for public release; still looking into that.

PR #3345, opened 2015-10-29
CSRF bypass based on header

priyank2005:
### Use case
We need a way to skip CSRF validation in case API is accessed by a backend system or mobile apps. To ensure that we have a method to specify skipping of CSRF validation using header in request

### Method

```javascript
module.exports.csrf = {
  grantTokenViaAjax: true,
  protectionEnabled: true,
  origin: '-',
  routesDisabled: '-',
  route: '/csrfToken',
  skipHeader: {
    enabled: true,
    headerName: 'x-csrf-skip-header',
    value: function() {
      return 'CSRF-SKIP'
    }
  }
};
```

If CSRF object is set as above then if request contains header as x-csrf-skip-header and value is CSRF-SKIP then request is properly processed.

### DIFF
--- a/lib/hooks/csrf/index.js
+++ b/lib/hooks/csrf/index.js
@@ -28,7 +28,14 @@ module.exports = function(sails) {
           protectionEnabled: true,
           origin: '-',
           routesDisabled: '-',
-          route: '/csrfToken'
+          route: '/csrfToken',
+          skipHeader: {
+            enabled: false,
+            headerName: 'x-csrf-skip-header',
+            value: function() {
+              return 'CSRF-SKIP'
+            }
+          }
         };
       }
       else if (sails.config.csrf === false) {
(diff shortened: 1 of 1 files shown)


sgress454:
Hi @priyank2005.  It seems like adding the ability to bypass CSRF protection via a header with a predetermined, set value would open up a security risk in certain situations (e.g. AJAX requests with permissive cross-origin settings).  And I'm hesitant to add any more to this hook than is absolutely necessary.  Is there anything preventing you from using the `/csrfToken` route to get the CSRF token, and send it with your browserless requests?

priyank2005:
Hi @sgress454, So basic idea here is not to allow a simple header string validation but providing a function that helps in validating header, that function can itself be hard enough so that spoofing request is not easy.

Use case is I want to have mobile application using APIs exposed in sails. These APIs are shared by website as well. And idea to use is to have a header value which has an RSA encrypted ttl string that will be verified for skipping CSRF.

mikermcneil:
> We need a way to skip CSRF validation in case API is accessed by a backend system or mobile apps. To ensure that we have a method to specify skipping of CSRF validation using header in request

Hi @priyank2005- as @sgress454 pointed out, this is the reason why we expose the `/csrfToken` endpoint.  You can have your mobile app or API send a request to that endpoint to obtain a CSRF token which you can then send with subsequent requests (which I think takes care of the issue which was necessitating the RSA token approach).  Hope that helps!

I'm going to close this PR in the interest of keeping on top of things.  If you have a related feature to propose in the future, please don't hesitate to do so (see the [new contribution guide for our updated process](https://github.com/balderdashy/sails/blob/master/CONTRIBUTING.md#v-proposing-features-and-enhancements))-- I appreciate you taking the time to put this together, and I think it'll help folks running into a similar use case in the future.

PR #3347, opened 2015-10-29
don't send stack track back for body parse errors when environment is production

nwhatt:
Not sure what if any tests should be updated for this.


sgress454:
Thanks @nwhatt!

RWOverdijk:
You're on fire @sgress454 :) One thing to note, shouldn't it move down 2 lines? Now the error doesn't get logged either.

sgress454:
Ah, good call.  I thought that was consistent with what we were doing in responses, but looks like we're logging the error there even in production (which makes sense).

RWOverdijk:
:) Still on fire though!

PR #3352, opened 2015-11-01
consle -> console

nha:


tjwebb:
This line is there so that console will still load in the case of slight misspelling. The subsequent command specifies the `console` command.

Issue #3363, opened 2015-11-11
Feature: Configurable API Rate Limiting


animedbz16:
I'm opening up this issue as a feature request since I didn't see a previously opened/closed issue and will submit a PR to the Roadmap Backlog section since this would be a nice to have feature if someone were to implement it, but this issue could be used to track further discussion for it.

I read from the Sails Docs in regard to DOS/DDOS (http://www.sailsjs.org/documentation/concepts/security/ddos), that there isn't much that can be done from an API perspective, however, I would like to suggest the implementation of a configurable API Rate Limit which could be used to prevent users from accessing the API too frequently.

This would ideally be configurable, such as allowing being able to configure X requests per second/minute/hour for each session, which if violated, Sails could intercept request and response to return a HTTP response of `429 Too Many Requests`, which could route to a new api response to allow users to change the behavior of the response.

Some additional examples of other APIs that implement this are here:
http://stackoverflow.com/questions/16022624/examples-of-http-api-rate-limiting-http-response-headers

weyert:
I think you can use existing Expres.js middleware for this in combination with Redis. That's what I am using for revoking tokens.

sailsbot:
Thanks for posting, @animedbz16.  I'm a repo bot-- nice to meet you!

It has been 30 days since there have been any updates or new comments on this page.  If this issue has been resolved, feel free to disregard the rest of this message.  On the other hand, if you are still waiting on a patch, please:

  + review our [contribution guide](https://github.com/balderdashy/sails/blob/master/CONTRIBUTING.md) to make sure this submission meets our criteria (only _verified bugs_ with documented features, please;  no questions, commentary, or bug reports about undocumented features or unofficial plugins)
  + create a new issue with the latest information, including updated version details with error messages, failing tests, and a link back to [the original issue]().  This allows GitHub to automatically create a back-reference for future visitors arriving from search engines.

Thanks so much for your help!

PR #3364, opened 2015-11-11
Add API Rate Limiting to ROADMAP

animedbz16:
More information about this feature request is tracked in #3363


sgress454:
Hi @animedbz16, this is interesting, but much better implement as a hook than a core feature.  We're trying really hard to fight feature bloat and move as much functionality as we can into separate modules.

PR #3372, opened 2015-11-17
Fix: Check for existence of association alias in pubsub

t3rminus:
As far as I can tell, `[association.alias]` is just creating an array with the alias value in it, and even `[undefined]` is truthy, so this will always pass.

Was this meant to check if `association.alias` is defined/truthy before using it in the accessor?


sgress454:
Right on @t3rminus, merging now.

PR #3380, opened 2015-11-20
Fix lint error

kt3k:
Hello
This change fixes lint errors and sets up `npm run lint` command in CI env.
Thanks,

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -30,7 +30,13 @@
   "supernew": true,
   
   // Allow backwards, node-dependency-style commas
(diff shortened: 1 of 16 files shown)


sgress454:
Thanks for pointing these out.  Not going to add a lint runner to our Travis tests because a) we use jshint and b) every time we do something like that, we regret it.  Tricky to rebase this one, so closing it and I'll go through with the jshint tool to fix what needs to be fixed.

kt3k:
OK. Thanks for considering anyway.

PR #3410, opened 2015-12-02
Add coverage status for the master branch

RomainLanz:
It's a great idea to add the coverage status of a project in the readme.

At the moment, this is hidden in the [Contributing Guide](https://github.com/balderdashy/sails/blob/master/CONTRIBUTING.md).

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,7 @@
 [![Gitter](https://badges.gitter.im/Join Chat.svg)](https://gitter.im/balderdashy/sails?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
 [![NPM version](https://badge.fury.io/js/sails.svg)](http://badge.fury.io/js/sails)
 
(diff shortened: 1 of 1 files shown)


mikermcneil:
Hi @RomainLanz - did the current placement of the coverage badge cause issues for you? I chose to put it in the contributing guide since it is only relevant for contributors.  When this issue first came up [two years ago](https://github.com/balderdashy/sails/pull/1317), I was all about the badges-- however, after trying this out, it became clear that, at least in our case, the coveralls badge causes confusion for new Sails users.  It also sets the precedent for adding all sorts of other badges and 3rd party plugins-- which can get out of hand pretty quickly and absorb too much of our core team's limited time and attentions.  

There's not necessarily a right or wrong answer here, and I respect the opinions of everyone on both sides.  That said, I'd like to keep this badge where it is.  Thanks for your feedback; I appreciate you taking the time to write this up.

PR #3453, opened 2015-12-29
Avoid duplicating work done by npm

lennym:
npm can resolve the path to a module from the top level app automatically with `require.main.require`, without having to first attempt to read the package.json, then the main property, then resolve again.

This also helps to support modules like (the newly released v1) `connect-mongo`, which do not define a "main" property in their package.json, and rely on defaulting to ./index.js.

### DIFF
--- a/lib/hooks/session/index.js
+++ b/lib/hooks/session/index.js
@@ -124,7 +124,7 @@ module.exports = function(app) {
           // or catch and return error:
           else {
 
-            var COULD_NOT_REQUIRE_CONNECT_ADAPTER_ERR = function (adapter, packagejson, e) {
+            var COULD_NOT_REQUIRE_CONNECT_ADAPTER_ERR = function (adapter, e) {
               var errMsg;
               if (e && typeof e === 'object' && e instanceof Error) {
                 errMsg = e.stack;
@@ -134,9 +134,6 @@ module.exports = function(app) {
               }
 
               var output = 'Could not load Connect session adapter :: ' + adapter + '\n';
-              if (packagejson && !packagejson.main) {
-                output+='(If this is your module, make sure that the module has a "main" configuration in its package.json file)';
-              }
               output+='\nError from adapter:\n'+ errMsg+'\n\n';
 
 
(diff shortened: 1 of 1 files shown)


lennym:
Note that the documentation at http://sailsjs.org/documentation/reference/configuration/sails-config-session may need updating if connect-mongo@1.x.x is to be supported, as it not longer constructs the mongo url automatically from `host`, `port` & `db` parameters.

mikermcneil:
@lennym dig it.  I think the only reason those tests are failing on node 0.12 is the timeout, which I just changed.  That said, we don't have proper tests using the mongo session adapter right now.  Do you have any ideas for how we could add some?

Thanks!

sgress454:
Working through this now.  I agree we can refactor this code, but unfortunately `require.main.require` won't work when you have Sails installed globally and do `sails lift` (it only works when doing `node app.js`).

Normally I'd just have you make the required change so that we can merge in your PR, but we're hustling trying to get v0.12 out ASAP so I'm going to close this and make the fix along with some other changes in that file today.  Thanks for pointing this out, we do appreciate the PR!

lennym:
No problem. I didn't consider the `sails lift` case because I'm used to using express, so `node index.js` is my goto approach but I totally understand that it's a primary use case for a lot of Sails users, so needs to work.

More generally though, I think it's probably important to avoid making these kind of assumptions about the install locations of modules where they can be loaded with `require`, especially with npm@3, but also with npm@2 if the app is not a top level dependency*.

I think the other place that I might look to do this is in this code - https://github.com/balderdashy/sails/blob/43fefa7860c9fa4edc07f6afbef0e2a6c1bf000b/lib/hooks/moduleloader/index.js#L399-L406 - which traverses the `./node_modules` to load hooks. I'd probably look to consider traversing the project's package.json dependencies, and require-ing each module's package.json to detect sails hooks etc. All of this has the same issue as here though in needing to be able to`require` from the context of the app and not sails.

\* Context: my current project has a microservices/service-oriented architecture in which the assorted services are dependencies of a parent app that handles orchestration and cross dependencies. This means that even with npm@2 it's possible that dependencies can end up being de-duped up a level to the top level app, and so don't exist in any one app's `./node_modules` directory, so these assumptions in sails have bitten us a couple of times already and meant that our deployment/installs can be a little fragile.

PR #3462, opened 2016-01-05
change error handling in res.jsonx

### LINKED ISSUE #3273: Blueprint errors inconsistant with 0.11 in 0.12.0-rc3
I am not sure if this was intended or not, but it seems like a bad regression if it was intended.  0.11 had a fairly good structure for presenting errors back to the client from blueprint calls, giving you the error code sails had, as well as raw info if you wanted that and so forth which I'm sure we are all familiar with.

0.12.0-rc3 seems to only be returning an error object with a "stack" property and a "message" property. There is information in the message property which I guess could be parsed out to find the same info as before, but it would be a lot nicer if we still had what we had in 0.11.x.

This is using mongo and sails-mongo adapter in case the issue is really related to those rather than sails/waterline, but since all I'm doing to get this change is to change sails from 0.11.2 to 0.12.0-rc3, I thought I'd put the issue here for now until/unless directed elsewhere.

fernandolguevara:
change error handling to fix #3273

### DIFF
--- a/lib/hooks/responses/index.js
+++ b/lib/hooks/responses/index.js
@@ -154,6 +154,33 @@ module.exports = function(sails) {
  */
 function _mixin_jsonx(req, res) {
 
+  function stringifyError(err) {
+    var plainObject = {};
+    Object.getOwnPropertyNames(err).forEach(function (key) {
+      plainObject[key] = err[key];
+    });
+    return JSON.stringify(plainObject);
+  };
+
+  function handleError(err){
+    var serializedErr,
+        jsonSerializedErr;
+
+      try {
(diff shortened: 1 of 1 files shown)


mikermcneil:
Thanks @fernandolguevara!

@robertd @robgrimball  Would you guys mind double-checking that this brings things into a backwards-compatible state with your 0.11 apps?  Once we're sure of that, we can add some additional tests that break on the current master branch, then merge this in (and at that point I'll go into sails-docs and ensure this is part of our [documented blueprint API](http://sailsjs.org/documentation/reference/blueprint-api/create) so we avoid backwards compatibility issues like this in the future)  Thanks!

PR #3476, opened 2016-01-08
Adding community SQL Server adapter. Recommending old MSSQL adapter f…

odnarb:
…or older sails versions as that package has not been updated in two years at the time of this commit (not including their README.md).


mikermcneil:
Thanks @odnarb.  Minor thing, looks like we'd want to change that to `<=0.9.x` since swelham's adapter is using the old version of the adapter spec from pre 0.10.  But realistically, we can just switch this out, since I think it's a fair assumption that folks using 0.9 are unlikely to be swapping database adapters without upgrading Sails core as well.

odnarb:
Ok, I wasn't sure. It's been a while since I used 0.9.16!
-brandon

RWOverdijk:
:+1: Looks good :)

mikermcneil:
Thanks!

Issue #3488, opened 2016-01-13
i18n hook should exposes __n() as well, not just __()


acis:
This is an issue when rendering server side templates from ajax calls.

Currently we're wrapping the call to jade's renderFile function and sending the request.__() and the locale in the options parameter.

````JS
var i18n = require('sails/node_modules/i18n');


function i18nRenderFile(req, path, options) {
    options.__ = req.__;
     //this doesn't work 
    options.__n = req.__n; 
    //this takes the default locale => bad 
    options.__n = i18n.__n; 
    options.locale = req.locale;

    return jade.renderFile(path, options);
}
````
 Seeing as the __n() function in not exposed in the i18n hook, it takes the default locale, which is undesirable.

**Comment by mikermcneil** [comment]
https://github.com/balderdashy/sails/pull/3489#issuecomment-171812260

**Comment by tswaters** [comment]
>  I need to understand more about the problem; in particular how this is something that needs to change in the core i18n hook, and the purpose of the __n function

The __n function is used to provide an additional numeric context to the translation to return a different string based on the number of items.... from the node-i18n docs,

```
  "%s kitty": {
    "one": "%s kitty",
    "other": "%s kittens"
  }
```

Due to the way i18n sets itself up, it will add it's entire API to `req` and `res.locals` automatically so one could use, say -- 

```
req.__n('%s kitty', 2); // 2 kittens
<% __n('%s kitty', 2) %> // in a view
```

However, setting it up on sails is done explicitly, and only `__` is added.  So under sails, many of the additional methods simply aren't available.

```
sails.__n('%s kitty', 2); // TypeError
```

The primary use case for using `sails.__` I find is for performing translations in a service.  Unfortunately, `__n` simply isn't registered, so it's not (easily) possible to perform numeric translations in a service.  A workaround is to pass `req` or `__n` into the service (which seems like a bad idea all around due to added complexity for tests).

I would argue that this is a bug... the current implementation simply does not provide all the methods that are available to a standard i18n object.

For older versions of sails, simply need to add `register: sails` to the configure option:

https://github.com/balderdashy/sails/blob/v0.12.13/lib/hooks/i18n/index.js#L78-L83

 In 1.0 I took a brief look at the code and it's changed to i18n-2, but still same idea, adding `register: sails` to the factory method would fix this:

https://github.com/balderdashy/sails/blob/v1.0.0-35/lib/hooks/i18n/index.js#L96-L101

Doing this will register the `resMethods` with sails:  
      __
[...]

PR #3489, opened 2016-01-13
expose __n() in i18n hook

acis:
fix for #3488

### DIFF
--- a/lib/hooks/i18n/index.js
+++ b/lib/hooks/i18n/index.js
@@ -84,6 +84,7 @@ module.exports = function(sails) {
 
 				// Expose global access to locale strings
 				sails.__ = i18n.__;
+				sails.__n = i18n.__n;
 			});
 
 			// Finally


mikermcneil:
Hi @acis- I just restarted the build for this PR (our timeout was too slow for 0.10, wasn't your fault it was failing).

I'm down to help here if I can, and https://github.com/balderdashy/sails/issues/3488 is a good start towards communicating your use case, but I need to understand more about the problem; in particular how this is something that needs to change in the core i18n hook, and the purpose of the `__n` function.  

Since exposing a new function is a change in behavior/usage rather than a bug fix, this would need to be submitted [as a proposal](https://github.com/balderdashy/sails/blob/master/CONTRIBUTING.md#v-proposing-features-and-enhancements).  Would you please review our updated contribution guide, then submit a proposal explaining your use case and how this approach would solve it?

Thanks!

acis:
will do, thanks!

mikermcneil:
@acis :+1:

mikermcneil:
btw @acis - would you mind closing this PR and https://github.com/balderdashy/sails/issues/3488 in the mean time to help us keep organized?  Thank you!

Issue #3502, opened 2016-01-19
Sails apps cannot use hooks published to npm as namespaced modules


lennym:
I was trying to load a custom hook from my own npm namespace by installing it from npm from `@lennym/sails-hook-sockets`, but because this installs into `./node_modules/@lennym/sails-hook-sockets` it is ignored by Sails.

I'd propose swapping some of this code (https://github.com/balderdashy/sails/blob/master/lib/hooks/moduleloader/index.js#L399-L407) for something that instead reads the app's package.json for dependencies, and then uses `require.resolve('<module>/package.json')` to read their `pacakge.json` files.

lennym:
Note: workaround is obviously to put the following in ./api/hooks/sockets.js:

```javascript
module.exports = require('@lennym/sails-hook-sockets');
```

It's just a little bit of a hack, and it's be nicer if it "just worked".

mikermcneil:
Thanks for the heads up @lennym- and for posting the workaround.  I was just looking at a similar issue re: adapters in the orm hook.  We'll look at it for hooks asap.  If anyone has time to submit a PR for this, that'd be awesome.

> I'd propose swapping some of this code (https://github.com/balderdashy/sails/blob/master/lib/hooks/moduleloader/index.js#L399-L407) for something that instead reads the app's package.json for dependencies, and then uses require.resolve('<module>/package.json') to read their pacakge.json files.

:+1:

lennym:
Thanks. I'll see if I can put together a PR today if I get some time.

Salakar:
@mikermcneil  @lennym  - I believe it's just the regex [here](https://github.com/balderdashy/sails/blob/master/lib/hooks/moduleloader/index.js#L436) that needs changing to:

```
/^(sails-hook-)|(@.+\/sails-hook-)/
```

Here's a regex101 of it: https://regex101.com/r/lC8sM4/1

lennym:
@Salakar That does not match my understanding from reading the code, which is that the code point you linked to will not be reached at all, because the module in question will never be recursed over, since it does not sit directly in a child directory of `./node_modules` with a package.json file at `./node_modules/<modulename>/package.json`. I will admit to not having debugged the code fully yet to establish this though.

[...]

PR #3514, opened 2016-01-21
[Patch] Update regex to parse hook names from module identities

lennym:
This allows the hook names to be correctly parsed from modules published to npm under a user's namespace/scope - e.g. `@user/sails-hook-name`

Follows discussion in #3502

### DIFF
--- a/lib/hooks/moduleloader/index.js
+++ b/lib/hooks/moduleloader/index.js
@@ -433,7 +433,7 @@ module.exports = function(sails) {
               }
               // Otherwise use the module name, with initial "sails-hook" stripped off if it exists
               else {
-                hookName = identity.match(/^sails-hook-/) ? identity.replace(/^sails-hook-/,'') : identity;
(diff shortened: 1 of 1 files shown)


Salakar:
I updated my comment on the issue, the regex can be reduced further to the below: 

```
/^(?:@.+\/)?(sails-hook-)/
```
See: https://regex101.com/r/lC8sM4/2

Sorry! :)

mikermcneil:
@lennym nice- this was the intended behavior (unless an `identity` is specified in the hook def, in which case that's used). Do you think you could tweak this pr to support using that if it's not undefined? Happy to help if I can-- this is what we need to be able to properly override core hooks via npm installed hooks (eg If you npm install an experimental version of sails-hook-orm-mongoose in your app that is fully compatible with blueprints and pubsub, then that hook needs to expose its identity as  'orm' ). @sgress454 and I can help with tests and docs on this if you don't have time (we were wanting to get this in anyway)

mikermcneil:
And @salakar, thanks for helping review!

sgress454:
Working on this now, adding some tests and should have it merged soon.

Issue #1177, opened 2013-12-07
Models in subfolders


ghost:
I created a controller and model "element" in the folder "Test" (generate by "sails generate test/element"). But the list of models hit "Test", which were assigned to all the necessary functions. "Element" (the path "Test.element") contained only the attributes of the model file. It is a normally? How can I reassign a functions for a "element"?
Sorry for my english

Issue #876, opened 2013-09-16
Sails.js Nested Folders For Controllers, Models, and Views


abbaabbbaabbbbbaa:
Please do so, that you can create subfolders in the folder 'controllers'. Example: api / controllers / tools / mails, api / controllers / tools / configs, api / controllers / tools / drivers. Where 'mails', 'configs', 'drivers' - is a controllers!

diosney:
@mikermcneil 

No offended :smile: 

Sorry for the very very late reply. I have to catch up with the updates since I moved on other projects, so I don't know exactly what is the status on this specific matter. (I don't wan)

No problem, I can understand your feeling but I think that a reasonable example is not needed here, just look at how the other frameworks have evolved to note that this is a reasonable and needed feature.

In my specific case, I had to build two large apps that needed to be extremely modular, with this I mean that the modules nesting was very deep due its nature, and having all related files contained in her own level and module helped a lot, I mean a lot, when plugin and unplug-in modules.

If at that time if sailsjs had this feature, I certainly had used it instead of plain expressjs.

I know maybe this is not the standard and common apps everyone builds, but indeed is a very useful feature that will bring several users to the framework.

Thanks

etiennegauthier:
I got a use case if anyone is interested. We are considering starting using sails as our main entry point for administration of everything. So in real terms, we will end up using sails on 3,4 maybe one day 10-15 databases/projects. I believe in our case, it is totally fair to expect having project organized in sub folders. Is there any workarounds ?

msqz:
You may access model (which exists in subdirectory)
`yourapp/api/models/yourmodule/SomeModel.js`
from controller (which also exists in subdirectory)
`yourapp/api/controllers/yourmodule/SomeModelController.js`
by using
`var SomeModel = sails.models['yourmodule/someorder'];`
`SomeModel.find()... whatever`
instead of just
`SomeModel.find()... whatever`
in the controller actions

sgress454:
@msqz That's not quite correct.  Currently Sails flattens the Models directory tree, but doesn't include folder names in the model identity.  So a model saved in `yourapp/api/models/yourmodule/SomeModel.js` would be accessed in Sails with (for example) `SomeModel.find()`.  See #3515 for an ongoing discussion of nested models/controllers.

Issue #1588, opened 2014-04-11
manage Models folder


Fr33maan:
Already existing in Issu #876 

Hello, the numbers of my models begin to increase (15 now, 30+ later and 50+ after).

I would like a good way to tidy up the models folder.

I have tried to do 'sails generate model folder/model' but i don't really understand whats happening : 
-i can call Folder.create
- i can NOT call Model.create

If i decide to manually create folders like /api/models/folder/Model.js, i can NOT do that : 
- Model.create

Someone know a good way to do what i want ? IMO sails would load models from models folder & sub folders.

Thanks

EDIT : maybe it is the flattenDirectories in lib/hooks/moduleloader/index.js wich must be configurable ?

PR #3515, opened 2016-01-21
(Feature) Nested Folders for Controller, Models, and Views

moisesrodriguez:
It's a feature is that has been requested here: [#876](https://github.com/balderdashy/sails/issues/876), [#1177](https://github.com/balderdashy/sails/issues/1177), [stackoverflow](http://stackoverflow.com/questions/27368946/create-subfolders-in-sails-js-api-controllers), [stackoverflow2](http://stackoverflow.com/questions/22059833/is-it-possible-to-group-controllers-in-sails-using-subfolders)


mikermcneil:
Hey @moisesrodriguez!  Thanks for bringing this up.

I took a look at the links you posted, and while they're helpful, we are definitely going to need a more detailed proposal in order for the discussion to continue here.  After reviewing each of those links, I was reminded of the widespread interest and thorough ideas/reasoning posted by people like @yomguithereal, @RWOverdijk  and @Hiro-Nakamura, which I appreciate.  While I don't personally understand I think they amply satisfy the requirement of understanding a use case, at least for nested controllers and/or nested views.

Yet I still don't have any single, cohesive picture of what these three different proposed features would look like for Sails users, let alone how their implementation would affect other existing usage in core.  Know what I mean?  It would be unrealistic to begin development of any of these three features before having a [clear vision](https://github.com/balderdashy/sails/blob/master/CONTRIBUTING.md#v-proposing-features-and-enhancements) of how it's going to work.  -- but even more importantly _why it's being requested_.


### Nested models
Before going any further, I should clarify that I still do not understand the use case for nested models.  I've seen this brought up a number of times over the years, usually in the context of apps with [50+ models](https://github.com/balderdashy/sails/issues/1588), and I understand the use case of being able to organize these modules.  Every time I've seen this come up, it's because someone is building a system which is working with a set of databases or models which is effectively dynamic; where the total number of models is growing as customers are added.  But this is not a good solution, because while it is quick as you're getting started, it's not scalable, and you will eventually corner yourself if you're successful.  A better solution is to use a single shared database (make the app truly multitenant)- or if that is not an option (i.e. regulatory requirements), then use branches or forks and deploy separate Sails processes for each customer.  Unless I'm missing something there, either of those two approaches removes the need for nested models.  I have yet to see a time where the topic of nested models comes up outside of this context, even in relatively abstract/complex projects such as a CMS built on top of Sails-- but of course I could be wrong!  So if anyone reading this has another use case for nested models, that _cannot be implemented without changing core_ please let me know.




### Is this actually subapps?

Is this actually subapps?  If not, then please skip ahead. But if so, can you think of a tangible, real-world use case that demands an aspect of subapps that isn't covered by the existing extensible hook system?  (for example, @sgress454 created [this hook](https://github.com/balderdashy/sails-hook-subapps) as an example of how you could build your own hook implementing subapps.  Yet even with that hook, we have yet to find a practical use case for it.  Again, I could be missing something, but I feel it's important to be transparent here.)

I remember using this kind of structure myself when I made a Grails plugin-- and I don't remember thinking it was particularly helpful.  In responding to this PR I spent about a half hour looking through the Grails plugin docs to refresh my memory, and even after that, I was not able to find even a single instance in that documentation explaining why the sub-app structure is useful.

I get the use case for _certain aspects_ of subapps: being able to key into various parts of the framework's lifecycle is useful, and grouping low-level logic by loose category is definitely a good idea (it's why we added hooks).  I get that other frameworks implement this-- I've even tried it out myself in Grails.  But that doesn't help us get to the bottom of _why people want this_.  I'm 100% willing to accept that this is something I just haven't come across the need for on any of our projects over the years-- it's just that without understanding the use case, and without seeing a compelling explanation of what would need to change in core to make that possible, it puts me in an impossible situation.



### Nested views

> If you refer to the section on "Providing Views, Templates and View resolution" in [this section of the Grails Plugin docs](https://grails.github.io/grails-doc/latest/guide/plugins.html), that's the kind of detail I'm looking for here, for each of the questions below.  

- What does it mean to have nested folders for views, exactly?
  - Would a default Sails app generated using `sails new` look any different?
  - If not, what would I have to do to use this feature?
  - Would this affect the usage of any existing documented feature in Sails?
    - e.g. would this affect any existing configuration in `sails.config` or add any new configuration?  If so, what would those changes look like?
  - Would this add any new methods, view helpers-- literally anything-- to programatic usage in a Sails app?
  - Would this affect the way views are currently loaded and looked up using `res.view()`?  What about when there is no first argument?  What about the view template cache loaded by express in production?


### Nested controllers

- What does it mean to have nested folders for controllers, exactly?
  - Would a default Sails app generated using `sails new` look any different?
  - If not, what would I have to do to use this feature?
  - Would this affect the usage of any existing documented feature in Sails?
    - e.g. would this affect any existing configuration in `sails.config` or add any new configuration?  If so, what would those changes look like?
  - Would this add any new methods, view helpers-- literally anything-- to programatic usage in a Sails app?
  - Would this affect the way controllers are currently routed to as explicit routes?
  - Would this affect blueprints?  Would blueprint actions take affect for nested controllers?
  - What about top-level configuration vs. nested configuration in general?
  - Would this support `target` (`{target: 'UserController.signup'}` ), `controller`+`action` (`{controller: 'UserController', action: 'signup' }`), and basic string route target syntax (`UserController.signup`), or would it work differently?
  - How would it affect `sails.controllers`?



### Implementation questions

Finally, answering these last questions would be helpful as well-- but without detailed answers to the interface questions above, it's obviously kind of hard to get into these.  That said, they all need to get answered before implementation can start- so in the interest of this discussion being able to forward in case it makes it to this step before I have time to review again, here goes:
- If you or I was to stay up a few nights this week and attempt to add these features to Sails, would I need to make any changes to the hook loader, configuration loader, or router in Sails core?  Would I need to change any core hooks like `moduleloader` or `controllers` or `views` or `orm`?  What about the core generators?
  - Tests?
  - Reference / conceptual docs?  What about the anatomy section?



Thanks again for bringing this up @moisesrodriguez, and sorry if this came across as me ragging on you.  I just want to get to the bottom of this once and for all, and this seems like as good an opportunity as any.

moisesrodriguez:
Hey @mikermcneil, the use case at least for why I would like Sails to support nested Models/Controllers is because my app has 60+ models and would like to organize them in folders. My app is actually multi-tenant and it would be great if Sails supported multi-tenancy with different database for each client as mentioned in this PR https://github.com/balderdashy/waterline/pull/787, but that is a different discussion.

In my case my models don't grow up as new clients are added. They all use the same models, it's just the complexity and many moving parts, that in this case I have many models.

Basically, what I would like to see is that if I add a folder in Models/Controllers e.g.

controllers
- Folder
  - controller1.js
  - controller2.js
- manyothercontrollers ...

models
- Folder
  - model1.js
  - model2.js
- manyothermodels ...

I could still access them like `sails.models.model1` and `sails.models.model2` and the controllers I could still access them like  `POST /controller1` and `POST /controller2` like you would normally do. Right now that does not work.

You ask if this would affect the way controllers and blueprints work. And the answer is no. Like I mention above I just want nested models for organization purposes. Not really looking for anything else to change.

I can see that if you organize them in folders it could potentially be confusing because you would then think that to access them, you would would have to do something like `sails.models.folder.model1` and `POST /folder/controller1`. Which I guess would add more complexity to implement nested folders.

>Would I need to make any changes to the hook loader, configuration loader, or router in Sails core? Would I need to change any core hooks like moduleloader or controllers or views or orm? What about the core generators?

I do not know the answers to that since don't know how the hook loader, configuration loader, etc work. But my guess is from what I'm describing is that probably not or not much.

Probably there are more use cases of how nested Models/Controllers could work, but in my specific case it's just being able to organize them. I understand I can have all my 60+ models in a flat structure and it's not the end of the world. But it would be nice if Sails supported nested models so blueprints would still work. Because there is really no point in organizing them right now in folders if blueprints, which are a very big feature in Sails, do not work and have to do many of the features Sails already does manually.

mikermcneil:
> my app has 60+ models and would like to organize them in folders....In my case my models don't grow up as new clients are added. They all use the same models, it's just the complexity and many moving parts, that in this case I have many models.

@moisesrodriguez Thanks for explaining that-- could you tell me more?  i.e. it would be helpful to see 3-4 of the model defs (model names+attributes+associations) to get a sense of how you're structuring things and if I can help.

Regardless of that though, the solution you described with a flat namespace for models and controllers, but allowing nested fetching would be a pretty simple change, and something I can get on board with (we use this strategy for client-side assets a lot and it works well to balance your ability to move stuff around w/o breaking your app with the need to provide some level of organization).  The only issues I can see right out of the gate are related to legacy code for federated controllers.  This would be a good opportunity to take care of that as well.

As for views, that does work slightly differently, but it's already supported:  (i.e. `res.view('foo/bar')` is different from `res.view('baz/bar')`). 

Would that help, you think?  If so, maybe we could start by adding experimental support for it on a branch, then get that working in your app, and then just see how it feels (I'll try it as well).  From there, assuming it works and feels right, then at that point we'll just need someone to add/update tests and docs, and we'll be able to merge it in.

Thanks again!

sgress454:
@moisesrodriguez @mikermcneil If you're just after organization and you're okay with a flat namespace, you can do that now.  You just need to use the `identity` property on your controllers and models.

```
// api/controllers/MyControllers/FooController.js
module.exports = {
   identity: 'Foo',
   hello: function(req, res) {return res.send('hello');}
};
```

```
// api/models/MyModels/Foo.js
module.exports = {
   identity: 'Foo',
   attributes: {
      name: 'string'
   }
}
```

This will give you a `Foo` model and controller equivalent to what you'd get by default with `api/controllers/FooController.js` and `api/models/Foo.js`, blueprints and all.  You'll be able to use the `Foo` global to reference your model in code as well.

moisesrodriguez:
@sgress454 @mikermcneil If that is the case, that is exactly what I want. Flat namespace is fine, it's just for organization purposes. I'll report back in case I run at any trouble. I'll keep this issue open for the time being in case someone wants to chip in with any use case they might have like the ones Mike mentioned above.

mikermcneil:
@sgress454 Thanks for bringing that up-- though @moisesrodriguez I should add that it's important to remember to set the identity, otherwise you could have issues.  If you run into anything after setting `identity`, please let us know here and we'll make sure we get that working.

It makes sense that this should work without having to necessarily set `identity`.  Right now there's still some leftover baggage from the various experiments with implementing feature requests for versioning-by-folders (e.g. `api/v1/`) and federated actions (e.g. `user/find.js`).  We later realized that these were a bit too strong of conventions to force in core-- but it left an imprint on the way controller modules were loaded.

I'd like to clean up the way we do this in the moduleloader -- specifically in `sails-build-dictionary` (not before 0.12 but soon).  You shouldn't have to set `identity` for this to work out of the box for controllers and models; I'd suggest it should be consistent with the behavior you get from files in `assets/js/` out of the box if you're using linker-injected client-side assets.  I realize this would introduce a potential conflict w/ duplicates, but we could just fail to load with a relevant error message in that case (I think that would preferable than the silent failure that a first-time user runs into when attempting to use a nested controller without specifying an explicit identity).

For example, here's a few scenarios to show what I'm getting at:


Normal situation:

```
api/controllers/pages/BrochureController.js
api/controllers/pages/AdminController.js
api/controllers/pages/DashboardController.js
api/controllers/endpoints/UserController.js
# yields:
# => BrochureController, AdminController, DashboardController, UserController

api/models/account/User.js
api/models/account/Role.js
api/models/account/Group.js
api/models/zoo/Zoo.js
api/models/zoo/Zookeeper.js
api/models/zoo/Specimen.js
api/models/taxonomy/Species.js
api/models/taxonomy/Genus.js
api/models/taxonomy/Family.js
api/models/taxonomy/Order.js
api/models/taxonomy/Phylum.js
# yields:
# => User, Role, Group, Zoo, Zookeeper, Specimen, Species, Genus, Family, Order, Phylum
```


If there is a duplicate filename, even in a different folder:
(note that the same thing applies if a duplicate identity or globalId is explicitly specified)

```
api/controllers/pages/BrochureController.js
api/controllers/pages/AdminController.js
api/controllers/pages/DashboardController.js
api/controllers/endpoints/AdminController.js
# => fails to load due to controller w/ duplicate auto-inferred identity (`admin`)


api/models/account/User.js
api/models/account/Role.js
api/models/account/Group.js
api/models/zoo/Zoo.js
api/models/zoo/Zookeeper.js
api/models/zoo/Specimen.js
api/models/taxonomy/Species.js
api/models/taxonomy/Genus.js
api/models/taxonomy/Family.js
api/models/taxonomy/Order.js
api/models/taxonomy/Phylum.js
api/models/checkout/Product.js
api/models/checkout/Order.js
api/models/checkout/PaymentMethod.js
# => fails to load due to model w/ duplicate auto-inferred identity (`order`)
```

sgress454:
@mikermcneil Sounds fine to me; this would be a breaking change because we _do_ support nested controllers, which we de-duplicate by having their identities include the folder name.  So to continue the above, the current, [documented](http://sailsjs.org/documentation/concepts/controllers/routing-to-controllers) situation is that 

```
api/controllers/pages/BrochureController.js
api/controllers/pages/AdminController.js
api/controllers/pages/DashboardController.js
api/controllers/endpoints/UserController.js
# yields
pages/BrochureController, pages/AdminController, pages/DashboardController, endpoints/UserController
```

and you can do things like this in `config/routes.js`:
```
"/some/route": "pages/BrochureController.someAction"
```

moisesrodriguez:
@mikermcneil @sgress454 so as you guys mentioned setting the identity works for organizing models/controllers in folders.

I'm all for what mike proposed without having to set the identity explicitly. 

>api/controllers/pages/BrochureController.js
api/controllers/pages/AdminController.js
api/controllers/pages/DashboardController.js
api/controllers/endpoints/UserController.js
 yields:
 => BrochureController, AdminController, DashboardController, UserController

>api/models/account/User.js
api/models/account/Role.js
api/models/account/Group.js
api/models/zoo/Zoo.js
api/models/zoo/Zookeeper.js
api/models/zoo/Specimen.js
api/models/taxonomy/Species.js
api/models/taxonomy/Genus.js
api/models/taxonomy/Family.js
api/models/taxonomy/Order.js
api/models/taxonomy/Phylum.js
yields:
 => User, Role, Group, Zoo, Zookeeper, Specimen, Species, Genus, Family, Order, Phylum

mikermcneil:
> as you guys mentioned setting the identity works for organizing models/controllers in folders
@moisesrodriguez glad to hear it!  Thanks for following up.


> I'm all for what mike proposed without having to set the identity explicitly.

Cool, I think it could be the right move-- the biggest potential issue is just w/ compatibility, particularly for blueprints.  I don't mind making a breaking change and putting this in v1; we'll just need to know what that migration would look like for everyone's existing Sails apps.

I think the next step here is to figure out exactly what is currently documented currently as far as the existing behavior of nested controllers -- particularly how they're written as route targets, and how they affect blueprints.  Because of other priorities for the v1 roadmap, I can't drive this proposal right now personally; but if anyone else reading has time/interest/need for this in the short term, I'll do what I can to participate in the discussion and help track things down.

sgress454:
Just an update, I realized while answering a related query that Sails does _not_ currently keep the directory names in the identities of nested models (in contrast to nested controllers).  So `api/models/foo/Bar.js` is exposed globally as simply `Bar`.  I recall the issue at the time being that we didn't know for sure _what_ people would want that model exposed as (but probably not `sails.models['foo/Bar']`).  So in the case of models it seems the only update would need to be a warning in case of duplicate identities (e.g. `api/models/foo/Bar.js` and `api/models/whiskey/Bar.js`).

Enethion:
So (if I understand it properly) if I create api with `sails generate api animals/mammals/lion` I will be able to reach route `localhost:1337/animals/mammals/lion` (== controller reached), but Model wont work (Controller automatically tries to reach `animals/mammals/lion` yet it's available as `lion`). When will this situation change? :smile:

Salakar:
@Enethion did you find a way around this? We need to build a cms of sorts to encapsulate multiple products we support so requires importing several different models, controllers and connections etc where there are duplicates identities, we want to namespace them by 'product' but just can't figure the best way around this. @sgress454 - maybe you have a suggestion / possible solution?

Enethion:
@Salakar I created issue #3731, but couldn't save it from closing. As I mentioned there: Setting identity "by hand" could work, but some model still would be overwrited. Only walkaround I found is creating Model+Controllers with dashes like `animals-mammals-lion`. By hand or by creating hook which will watch changes in directories/files and do copying+renaming automatically.

boris-lapouga:
Another case supporting the thread maker:
We're developing a huge (country wise) medical system that includes hundreds of models and controllers, unfortunately Sails was picked at the very beginning, with the inability of Sails to adapt to any folder structure that suitable for development convenience, code readability and rapid development I'm thinking of getting rid of the system now, though we're too far away into the development.

That could be an amazing showcase project for the author, but I see that no argument can convince him :-)

Our developers have picked a folder structure that suits their needs for architectural expansion, we don't have models and controllers folders, we're having an /app folder inside of /api folder, inside of an /app we have components defined as folders, each component contains one controller and many models bound to it, while I understand that a single controller is bad practice for many models, still, not having an ability to teach Sails how to work with this structure forces us each save of our changes (Ctrl+S) run another hotkey (Alt+S) that runs .sh script which pulls all files from the folders and lays out all models in /models folder and controllers in /controllers folder respectively, to just make sure Sails will work with it, which is a huge overhead for us and a point of failure during deployments to different stages of our infrastructure.

I'd really appreciate just a hook from the system where I could build my own node plugin that will tell Sails where to look for models, controllers and how they are structured. This is not a must to be defined in Sails core itself, a few hooks would help a lot so each and everyone can define their own structure, and there should be only one argument that actually works: We'd like, according to our project size, define the folder structure that suits our needs as users of this package.

Thanks

LiorArbel:
@boris-lapouga summarized exactly my team's project problem with sails.

I have spent the last three days looking for any way to change the directory structure to work even remotely in the component style (plugins, sub apps, hmvc etc...) and have come up with nothing really satisfactory - every issue/pull request usually contained a hack-ish workaround or a grand debate whether you should even do that, and either way in the end they all got closed. 

Seeing as this pull request is still open and @boris-lapouga 's question is the latest unanswered one I have found on this "debate", I would be really glad to hear your response. 

P.S. My team decides tomorrow what technology we should use for our new server, and this issue is going to be a major drawback for them, as we are a team of 11 programmers working on the same huge server and also we are really harsh on conventions.

I just feel I have to strongly clarify that if we choose sails, our first task will be to write exactly boris' code all over again (maybe like many more silent people) or just write plain Express code while losing all those great great advantages of sails that I love so much.

mikermcneil:
@Salakar @LiorArbel @boris-lapouga @Enethion Hey guys, I apologize for not getting back earlier-- just noticed your comments.  Responses inline below:

> I'd really appreciate just a hook from the system where I could build my own node plugin that will tell Sails where to look for models, controllers and how they are structured.

I think this is a great idea.  You can override the `moduleloader` hook for this purpose (although I realize that can be a bit intimidating since there's a lot going on in there).  I wonder if an example demonstrating how to do that would help out here?


> Our developers have picked a folder structure that suits their needs for architectural expansion, we don't have models and controllers folders, we're having an /app folder inside of /api folder, inside of an /app we have components defined as folders, each component contains one controller and many models bound to it, while I understand that a single controller is bad practice for many models, still, not having an ability to teach Sails how to work with this structure forces us each save of our changes (Ctrl+S) run another hotkey (Alt+S) that runs .sh script which pulls all files from the folders and lays out all models in /models folder and controllers in /controllers folder respectively, to just make sure Sails will work with it, which is a huge overhead for us and a point of failure during deployments to different stages of our infrastructure.

@boris-lapouga That's unfortunate, and it does sound complicated.  To take a step back-- is the main goal here isolating who works on what files?  If so, I completely understand what you're after (and can relate to the need)-- or if not, please correct me so I can make sure I understand.

> I have spent the last three days looking for any way to change the directory structure to work even remotely in the component style (plugins, sub apps, hmvc etc...) and have come up with nothing really satisfactory - every issue/pull request usually contained a hack-ish workaround or a grand debate whether you should even do that, and either way in the end they all got closed.

@LiorArbel Would you mind making a quick example repo demonstrating the folder structure you'd like Sails to support for your app?  The actual code inside can be fake, it'd just be helpful to see an example of the files and where they live (eg. "CatController", "Dog", etc).

--------------------

Like I mentioned above, a lot of the flexibility y'all have mentioned can be accomplished by overriding the moduleloader hook.  But I realize that this is easier said than done.

So I reckon the way I can be most helpful here is probably by putting together an example of how to do that (and cleaning up anything I notice along the way that would make it harder/less clean to do).

Thanks for taking the time to write up detailed responses, and I look forward to helping out further if I can!

boris-lapouga:
Hey,

Thanks for response. Will there be any follow up with examples?

I wonder if it would be possible from your side, just to make an optional way of hooking into moduleloader and give it some JSON based file structure to expect, if not provided, fallback to current functionality.

Even simpler, you can just use glob to recursively parse some provided directory, say /api/app, an by matching either explicitly set regex for Model, Controller, Routes, Services or just hardcode to look for XModel.js, YController.js, ZService.js, XXRoutes.js, then either lay them down in linear structure like it does now or keep pointers to files to load when needed.

Issue #2665, opened 2015-02-18
Binding services to the module context breaks static properties on methods.


natecavanaugh:
Hi there,
This was working in 10.4, but in 11, as of 97861aa30da56eef2907516010ac31157e8c3abe there is a use case that is broken.
Let's say we have `EmailService.js`
```
var EmailService = function(){
// Do some stuff
};
EmailService.foo = 'Some Value';
module.exports = EmailService;
```

When the module loader is calling _.bindAll, it's replacing all methods on the object with a "bound" version, but it completely wipes out any existing properties on that function.

It could be all well and good to say "Don't export functions", but 1. it does it to every function on the object, not just the exported one (so if `EmailService.foo` were a function, and had properties, those would be wiped out as well), and 2. we may be exporting a third party library, where we don't control the kind of object that's being exported.

Since `_.bindAll` is destructive, I would suggest making this something the service can opt into or control themselves.

sgress454:
> When the module loader is calling _.bindAll, it's replacing all methods on the object with a "bound" version, but it completely wipes out any existing properties on that function.

That would be terrible!  Fortunately the exact code you posted works fine for me on a fresh install of v0.11 (tried `EmailService.foo` in `sails console`, as well as `console.log(EmailService.foo)` inside of the function), so I suspect the issue is elsewhere.

natecavanaugh:
Sorry, for some reason I thought I had tested the example thoroughly, but testing it, you are correct. Where it fails is the following example:
```
var EmailService = function(){
// Do some stuff
};
EmailService.SomeFn = function() {
	// Do some other stuff
};
EmailService.SomeFn.foo = 'Some Value';
module.exports = EmailService;
```

So basically, if the property on the function is also a function that contains properties, those props are lost.

Are you able to reproduce the same issue?

Thanks again for looking at it :)

sgress454:
Ok, that I can reproduce, and it makes sense.  I suppose this is more an issue with Lodash's `_.bindAll` function than with Sails core, but that's neither here nor there.  Can you work around it?  Is it imperative that your service functions have their own properties?  Or could you just store those properties in a hash at the top level of the service, or better yet in a private var that doesn't get exported at all?

natecavanaugh:
Thanks for the response, but I don't know if I can work around it for all of my scenarios, as it's a third party lib that's exposing utility methods on a function that also does some work on it's own.
[...]

PR #3536, opened 2016-01-31
[Proposal] Don't bind services to module context

mikermcneil:
This behavior will stay as is for 0.12, but I'd like to tentatively plan to remove the automatic binding of context in services for Sails v1.0.  I don't have time right now to spec this further, however I don't think this one will be a whole lot of work to spec out.  The biggest thing is to put together a list of links to the places in the docs this will need to change (or get added).
# See [original issue](https://github.com/balderdashy/sails/issues/2665) from @natecavanaugh for more info.  Included pertinent info below for context:

Hi there,
This was working in 10.4, but in 11, as of 97861aa30da56eef2907516010ac31157e8c3abe there is a use case that is broken.
Let's say we have `EmailService.js`

```
var EmailService = function(){
// Do some stuff
};
EmailService.foo = 'Some Value';
module.exports = EmailService;
```

When the module loader is calling _.bindAll, it's replacing all methods on the object with a "bound" version, but it completely wipes out any existing properties on that function.

It could be all well and good to say "Don't export functions", but 1. it does it to every function on the object, not just the exported one (so if `EmailService.foo` were a function, and had properties, those would be wiped out as well), and 2. we may be exporting a third party library, where we don't control the kind of object that's being exported.

Since `_.bindAll` is destructive, I would suggest making this something the service can opt into or control themselves.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -52,7 +52,7 @@ Change                                           | Proposal                     
  :---------------------------------------------- | :------------------------------------------------------------------------------------ | :----------------------------------------------------------------------------------------------------------
  Generate `test/` folder in new Sails apps       | [#2499](https://github.com/balderdashy/sails/pull/2499#issuecomment-171556544)        | Generate a generic setup for mocha tests in all new Sails apps.  Originally suggested by [@jedd-ahyoung](https://github.com/jedd-ahyoung).
  `sails.getRouteAddress()`                       | [#3402](https://github.com/balderdashy/sails/issues/3402#issuecomment-167137610)   | Given a route target, return the route address configured in the app's explicit routes.
-| Don't bind context in app's services           | []() | Don't bind context (`this`) in modules loaded from `api/services/`
+| Don't bind context in app's services           | [#3536](https://github.com/balderdashy/sails/pull/3536) | Don't bind context (`this`) in modules loaded from `api/services/`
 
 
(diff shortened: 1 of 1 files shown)


cage1618:
I get the same issue, hope to resolve as soon as possible!

sgress454:
I'd propose that we leave the binding on by default, and allow a service to opt-out via some configuration.  The reason being that we expect Sails apps to be able to be fully functional with globals turned off, and with globals turned off there's no other way besides `_.bind` to provide modules with access to the `sails` object which contains configuration and other goodies.  Also, all of the other modules (models, controllers, policies, responses, adapters, and hooks) are bound to Sails, so not doing so with services would be inconsistent.

In any case, the only time this isn't easily worked around right now is if you have a service like:

```
module.exports = require('someThirdPartyModule');
```

(see the snippet at the bottom of [this comment](https://github.com/balderdashy/sails/issues/2665#issuecomment-172420518) for workaround) 

But if mine is the minority opinion, I have no problem with this proposal as-is...

mikermcneil:
I haven't seen a whole lot of interest on this one in the form of folks gathering up docs, so I'm going to go ahead and close it.   Sails v1 introduces **helpers**, stateless functions, basically, which take the place of services (services are still supported for convenience, but `api/services/` will not be generated in new Sails apps).

PR #3540, opened 2016-02-03
Update ROADMAP.md

jasalazar:
Call controller's actions directly from views.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -82,5 +82,6 @@ Feature                                          | Summary
  Manual migrations in Sails CLI                  | For production environments it would be nice to have a save/secure command that creates the db automatically for you; e.g. a `sails migrate` or `sails create-db` command.  See [sails-migrations](https://github.com/BlueHotDog/sails-migrations) and [sails-db-migrate](https://github.com/building5/sails-db-migrate) for inspiration.  We should begin by contributing and using one or both of these modules in production in order to refine them further into a full fledged proposal (the Sails core team is using sails-migrations currently).  Originally suggested by [@globegitter](https://github.com/Globegitter).
  Wildcard action policies                        | Instead of only having one global action policy `'*'` it would be nice if we could define policies for a specific action in all controllers: `'*/destroy': ['isOwner']` or something similar.  Originally suggested by [@ProLoser](https://github.com/ProLoser).
  SPDY/HTTP2 protocol support                     | See https://github.com/balderdashy/sails/issues/80 for background.
+ Call controller's actions from the view        | Allow to call actions directly from views. [See this issue for more details](https://github.com/balderdashy/sails/issues/3535)
 
 


mikermcneil:
> [**Please do not submit a pull request _adding_ to this section.**](https://github.com/jasalazar/sails/blob/31b3444fe1dc023028e5a1d578aeb701b0d00b9f/ROADMAP.md#pending-proposals)
@jasalazar would you mind rereading the section on [submitting a proposal in the contribution guide?](https://github.com/balderdashy/sails/blob/master/CONTRIBUTING.md#v-proposing-features-and-enhancements).  We revamped the process recently.  Thanks!

mikermcneil:
(@jasalazar also I think what you're asking about might actually be supported in 0.12-- not sure, but see https://github.com/balderdashy/sails-docs/blob/0.12/reference/application/sails.getUrlFor.md)

mikermcneil:
e.g.

```html
<form action=“<%= sails.getRouteFor(‘MeController.signup’).url %>” method=“<%= sails.getRouteFor(‘MeController.signup’).method %>”>
```

mikermcneil:
@jasalazar that aside though, I think what you probably want to do is use AJAX.  In general, the recommended process for communicating with your backend controllers from your views in Sails is using AJAX requests (whether that's via something like `$.ajax()` or `io.socket`).

jasalazar:
Thanks for the answer @mikermcneil.

What i really wanted was find a way to call controller's actions from the client without having to send an `$.ajax()` request.

I also am interested in finding a way to avoid showing `$.ajax()` or `io.socket` calls on the client side.

For the moment, for me it works having a JS file with a function which calls an `$.ajax()` to a route set on `sails.config.routes` just like this:

On the client side:

 `<select id="a" onchange="foo()"><option value="<%= global.id %>"></option></select>`

On `myFile.js`:

`function foo(){`
`id_option = $("#a").val();`
`$.ajax({`
`    type: "GET",`
`    url: '/route',`
`    dataType: 'json',`
            `data: {`
             `   'id': id_option`
            `},`
            `contentType: 'application/json',`
            `async: true,`
            `cache: false`
        `}).done(function(data){`
`#dosomething`
`)};`
`}`

On my `config/routes.js` file:

`'GET /route': 'MeController.action'`

On my controller i am calling a method located on my `Me` model, which sends an `$.ajax()` request to my API, which is located outside Sails, on the cloud.

Also, it would be fine to be able to hide my .js files so they wont be seen via browser.


PS: I'm sorry i didn't understand too well the correct way to send a proposal.

mikermcneil:
> PS: I'm sorry i didn't understand too well the correct way to send a proposal.

@jasalazar no problem-- thanks for taking the time to add some more information :+1: 

When you say:
> I also am interested in finding a way to avoid showing $.ajax() or io.socket calls on the client side.
Do you mean avoid showing the URLs and methods?

For front-end code obfuscation in general, you might check out something like [`grunt-jscrambler`](https://github.com/jscrambler/grunt-jscrambler).  Although depending on your requirements, the built-in minification support in Sails might be good enough (I've never needed more than that for front-end code).  Also remember that no matter what you do with your front-end code, your backend endpoints need to be protected (somebody could just blast Postman requests at any endpoint they like).

jasalazar:
Thanks again @mikermcneil for your response and guide through my proccess of learning and mastering Sails.

What i meant to say, is i want to avoid showing everything an ajax call includes, whether be URL's or methods, everything!

What would you suggest for protecting my backend endpoints?

Regards.

mikermcneil:
@jasalazar to protect your endpoints, you'll want to use policies (and in some cases, custom code in your actions).  I'd suggest checking out https://courses.platzi.com/courses/develop-apps-sails-js/ and https://github.com/sails101/basic-login.  Hope that helps!

PR #3541, opened 2016-02-03
Support passing configuration options to the default body parser

lennym:
I'd like to be able to set config options for the default ('skipper') body parser without having to explicitly depend on it, and pass it as a body-parser function in my app.

e.g. in config/hooks.js

```javascript
module.exports.http = {
    ...
    middleware: {
        bodyParser: {
            options: {
              limit:'10mb'
            }
        }
    }
    ...
}

As it is, I am forced to pass `fn: require('skipper')` and add skipper to my dependencies, which is a bit longwinded to simply pass an option.

By defaulting the `fn` option to skipper if the bodyParser config is an object then Sails can support passing options only to the default body parser.

### DIFF
--- a/lib/hooks/http/middleware/defaults.js
+++ b/lib/hooks/http/middleware/defaults.js
@@ -109,7 +109,7 @@ module.exports = function(sails, app) {
         // middleware provided with options
         if (typeof conf === 'object') {
           opts = conf.options || opts;
-          fn = conf.fn;
+          fn = conf.fn || require('skipper');
           return fn(opts);
         }
         // middleware function defined directly


lennym:
Actually, looking closer, the entire code block at https://github.com/balderdashy/sails/blob/master/lib/hooks/http/middleware/defaults.js#L107-L155 to read from `sails.config.http.middleware.bodyParser` is effectively a no-op, since if that property resolves to anything other than `undefined` then the `_.defaults` call at https://github.com/balderdashy/sails/blob/master/lib/hooks/http/middleware/defaults.js#L21 will simply use the value from the config.

In particular, this means that setting any config to `sails.config.http.middleware.bodyParser` that isn't an actual body parser function will cause requests to error/hang. The code is simply wildly misleading.

PR #3546, opened 2016-02-04
Before adding a userhook to be loaded, check that it was not marked as 'not' to load via config

luislobo:
It fixes https://github.com/balderdashy/sails-hook-sockets/issues/16, where the sails-hook-sockets is disabled via config, but since it is a module, it is being loaded via the userhook method

### DIFF
--- a/lib/hooks/moduleloader/index.js
+++ b/lib/hooks/moduleloader/index.js
@@ -479,6 +479,7 @@ module.exports = function(sails) {
                 memo[hookName] = hook;
               }
             }
+            return memo;
           }, {}));
 
           return bindToSails(cb)(null, hooks);
--- a/lib/hooks/moduleloader/index.js
+++ b/lib/hooks/moduleloader/index.js
@@ -475,9 +475,10 @@ module.exports = function(sails) {
               hook.configKey = (sails.config.installedHooks && sails.config.installedHooks[identity] && sails.config.installedHooks[identity].configKey) || hookName;
 
               // Add this to the list of hooks to load
-              memo[hookName] = hook;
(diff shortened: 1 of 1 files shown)


sgress454:
Hi @luislobo -- thanks for tracking this down.  I think what we need to do is patch the [version in the stable branch](https://github.com/balderdashy/sails/blob/stable/lib/hooks/moduleloader/index.js) with the fix that's [already there on master](https://github.com/luislobo/sails/blob/master/lib/hooks/moduleloader/index.js#L448).

luislobo:
well, yes, do you want me to just patch it? I mean, I generated the PR already... Sorry if  I don't get your idea :grin:

luislobo:
Oh, I see what you mean, put both conditions on the same line... right?

luislobo:
Well, I've checked and that fix actually is enough! I've compared both sails.config.hooks[hookName] and sails.hooks[hookName] and both contain the same value when setting sockets : false.
I think that then, this PR is not necessary.
We could change it to ```if (sails.config.hooks[hookName] === false || sails.hooks[hookName] === false) {``` to be very specific...
I've tested a new app using the current master branch and it does not fail.

sgress454:
@luislobo Sorry if I wasn't clear.  This problem is already fixed on master; no patch is necessary.  The issue balderdashy/sails-hook-sockets#16 is happening on the current _stable_ version of Sails (v0.11.x), which can be patched by making a PR to the `stable` branch.  It would be nice to have 0.11.x fixed as well, even though we're about to release 0.12.

luislobo:
New PR here: https://github.com/balderdashy/sails/pull/3550

PR #3552, opened 2016-02-07
fixed sails.sockets.id() is deprecated; use sails.sockets.getId() warning msg

masitko:
Sails 0.12.0 deprecated sails.sockets.id() method. 
One of the core hooks pubsub still uses this method and triggers a warning
```
sails.sockets.id() is deprecated; use sails.sockets.getId().
```
This patch fixes the issue.


mikermcneil:
Thanks @masitko!

sgress454:
Fixes #3563

ljacques:
gg! (is a French term meaning great)

PR #3569, opened 2016-02-11
Do not merge!

sgress454:
Using this PR to test some possible fixes to the issues with the Travis tests.  Step one:  add a timeout on all calls to `.lower()`.


mikermcneil:
@sgress454 perhaps it's time to consider just switching to circle ci.  Have they been consistently passing there?  Hell, I'm fine with them consistently failing.  As long as the test server is doing _something_ consistent.

sgress454:
@mikermcneil argh, I don't know.  There are fewer random test failures on CircleCI to be sure--but it is also running 1/4 of the tests that Travis runs, since Travis does four Node versions at once.

mikermcneil:
@sgress454 I think perhaps the various great test roundups of 2015 and 2016 have worked (haven't seen any ghost in the shell stuff on travis anymore recently).  I think the randomness of it was likely due to the timing of when `sails.lower()` was firing its user-provided callback (and also some of the tests not waiting for that callback before proceeding).  I'm going to go ahead and close this, but if this stuff rears its ugly head again, well.... we know where to find it.

PR #3572, opened 2016-02-11
Added ability to specify which columns to select in the find blueprint

jwenz723:
Added ability to specify which columns to select in the find blueprint. With this, columns can now be specified in the URL when calling /:Model by simply adding the "columns" parameter. The columns should be separated by a comma. I.e. /:Model?columns=column1,column2

### DIFF
--- a/lib/hooks/blueprints/actionUtil.js
+++ b/lib/hooks/blueprints/actionUtil.js
@@ -206,7 +206,7 @@ var actionUtil = {
 
     // Allow customizable blacklist for params NOT to include as criteria.
     req.options.criteria = req.options.criteria || {};
-    req.options.criteria.blacklist = req.options.criteria.blacklist || ['limit', 'skip', 'sort', 'populate'];
+    req.options.criteria.blacklist = req.options.criteria.blacklist || ['limit', 'skip', 'sort', 'populate', 'columns'];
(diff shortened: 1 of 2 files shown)


jwenz723:
I also created this as an npm module that can be added on top of sailsjs.  This can be seen here: https://bitbucket.org/jwenz723/sails-hook-blueprint-findcolumns/overview

sgress454:
Hi @jwenz723, thanks for the PR.  A couple of points:

 * We have a new [contribution guide for proposing features and enhancements](https://github.com/balderdashy/sails/blob/master/CONTRIBUTING.md#v-proposing-features-and-enhancements) that we ask you to check out to help streamline the process of reviewing new code.  You'd need to write up a proposal to make the case for adding this to the Sails core feature set.
 * Creating a hook is almost always a better idea than attempting to add features to core (because we're trying to keep the core as lean as possible).  So your work might be done already.
 * It looks like this particular feature relies on a private/undocumented `select` criteria property in Waterline, so it's unlikely that it would be merged into core anyway.  You'd need to rewrite the code to not rely on that property.  As a hook, it's fine, you just need to be aware that the private API could change at any time and render the hook unusable (so basically "caveat emptor" for your users!)

PR #3599, opened 2016-02-25
Ensure req._sails is always set

### LINKED ISSUE #3598: Possible that req._sails does not exist depending on configuration
I've been trying to debug an issue with the default response handlers (`negotiate`, `serverError` etc) not working, and have discovered that it's a result of depending on `req._sails` being heavily depended on in order to access sails.log.

The cause of this issue being that in our app we have disabled the request hook in order to prevent the parsing of `req.params.all()` since the repeated invocation of `_.cloneDeep` contained within the middleware to set that up caused catastrophic performance issues with moderately large json bodies (Context: response times went from > 2 minutes to < 10s when the hook was disabled).

Since `req._sails` is heavily depended on in a number of places, it would be helpful if it were not so easily lost when disabling hooks which are not required.

I'd propose looking to create a single hook for the creation of the `req._sails` property - and potentially any other properties which are potentially depended on in sails code (as opposed to users' code) - that is accompanied by a warning that sails itself may stop working if it is disabled. Alternatively, some method by which to disable the parsing for `req.params.all()` independently since it has such severe performance implications that its removal is likely to be necessary.

lennym:
Certain parts of the codebase - in particular the built is response functions (e.g. `res.serverError` etc) access `req._sails` to gain access to log functions. However, the setting of `req._sails` is currently bundled with the "request" hook, which contains potentially undesirable code - in particular around parsing `req.params.all()`, which can have a severe performance impact in certain situations - so may be disabled.

By setting `req._sails` by default on all requests we can ensure that the built in response helpers are usable indepenently of what intermediate hooks and middleware have loaded.

Fixes #3598

### DIFF
--- a/lib/hooks/http/initialize.js
+++ b/lib/hooks/http/initialize.js
@@ -99,6 +99,11 @@ module.exports = function(sails) {
             });
         }
 
+        app.use(function (req, res, next) {
+            req._sails = sails;
+            next();
+        });
+
         // When Sails binds routes, bind them to the internal Express router
         sails.on('router:bind', function(route) {
 
--- a/test/integration/middleware.sails.test.js
+++ b/test/integration/middleware.sails.test.js
@@ -1,0 +1,59 @@
+var request = require('request');
+var Sails = require('../../lib').Sails;
+var assert = require('assert');
+
+describe('middleware :: ', function() {
+
+  describe('sails :: ', function() {
+
+    describe('http requests :: ', function() {
+
+      var sid;
+
+      // Lift a Sails instance in production mode
+      var app = Sails();
+      before(function (done){
+        app.lift({
+          globals: false,
+          port: 1535,
+          environment: 'development',
+          log: {level: 'silent'},
+          session: {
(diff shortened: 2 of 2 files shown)


lennym:
Not quite sure why the CI is failing. It's a timeout and not in any code that's been touched in the PR, so I suspect it may be a transient error.

mikermcneil:
> Not quite sure why the CI is failing. It's a timeout and not in any code that's been touched in the PR, so I suspect it may be a transient error.

Yeah it's something we've been struggling with.  Locally, it doesn't seem to be an issue

mikermcneil:
@lennym love it!  Only thing I'd ask is that we find a way to make sure this is also true for virtual requests coming from the sockets hook or e.g. `sails.request()`.  It may already be taken care of, I'm not sure

mikermcneil:
@lennym actually, disregard that-- for virtual requests, I set it up to run before literally everything else apparently: https://github.com/balderdashy/sails/blob/master/lib/router/index.js#L184

Thanks man!

PR #3601, opened 2016-02-25
[fixes #3592] allow sails.config.http.middleware.bodyParser config to work

### LINKED ISSUE #3592: Cannot load custom bodyParser config from `sails.config.http.middleware.bodyParser`
To reproduce:

1. In `config/http.js`, uncomment example middleware config (at minimum the `middleware: {` property and closing `}`, and the `bodyParser: require('skipper')` line.
2. `sails lift`
3. Make any request (e.g. to the homepage) and notice that the request hangs.

The issue is that in `defaults.js` in the HTTP hook code we [go to great lengths to allow overriding of the body parser via `sails.config.http.middleware.bodyParser`](https://github.com/balderdashy/sails/blob/master/lib/hooks/http/middleware/defaults.js#L116-L154), but since all of that code is wrapped in `_.defaults()`, the result of the self-calling function is never actually used, because a value already exists for the `bodyParser` key.  In the example above, the loaded body parser would be the result of `require('skipper')`, which is not itself valid (it needs to be called w/ options).  Furthermore, doing:

```
bodyParser: require('skipper')(<some options>)
```

doesn't work, because the self-calling function in `defaults.js` would then call _that_ resulting function, resulting in an error.  

Currently the only way to customize the bodyParser is via `sails.config.http.bodyParser` (not `middleware.bodyParser`).  We should probably keep this for backwards-compatibility (with a deprecation warning), but in the self-calling function in `defaults.js`, just return early if `sails.config.http.middleware.bodyParser` is defined.

sgress454:
This basically removes most of the code from the `bodyParser` property of the object in `hooks/http/middleware/defaults.js`.  I think this code was written when Skipper was just getting started and needed a bit more handholding, but it probably hasn't worked for a long time.  It's _supposed_ to take the value of `sails.config.http.middleware.bodyParser`, validate it, and use it to get a body parser middleware function to use in the app.  But because all of the code in `defaults.js` is wrapped in a big `_.defaults(sails.config.http.middleware, {...}`, the function that gets returned is never used, since there's already something in `sails.config.http.middleware.bodyParser`.  The catch 22 is, an Express middleware function will _not_ pass the validation in `defaults.js`, so an error will be thrown and Sails won't lift--but if you put in the kind of value that _will_ pass validation, then that value (and not an Express middleware function) will get used as the body parser, and all requests will hang.

A corollary to this PR would be to change the example in `config/http.js` from:
```
bodyParser: require('skipper')
```
to
```
bodyParser: require('skipper')({maxTimeToBuffer: 24400})
```
(or some other options; just anything that demonstrates proper usage)

You _can_ currently specify a custom config by using `sails.config.http.bodyParser`, so we should keep and deprecate that (undocumented) usage.  But at this point our http hook is solid enough that it's safe to have `middleware.bodyParser` be the same as all the other middleware config:  you just give it a function with `req`, `res` and `next`.

### DIFF
--- a/lib/hooks/http/middleware/defaults.js
+++ b/lib/hooks/http/middleware/defaults.js
@@ -111,58 +111,6 @@ module.exports = function(sails, app) {
         return undefined;
       }
 
-      // Handle new middleware config:
-      ////////////////////////////////////////////////////////
-      var conf = sails.config.http.middleware.bodyParser;
-      if (conf) {
-        // middleware provided with options
-        if (typeof conf === 'object') {
-          opts = conf.options || opts;
-          fn = conf.fn;
-          return fn(opts);
-        }
-        // middleware function defined directly
-        else if (typeof conf === 'function') {
-          fn = conf;
-          var configuredFn;
-
-          try {
-            configuredFn = fn(opts);
-          }
-          catch (e) {
-            sails.on('lifted', function () {
-              sails.log.error(e);
-              sails.log.blank();
-              sails.log.error('Encountered an error when trying to use configured bodyParser.');
-              sails.log.error('Usually, this means that it was installed incorrectly.');
-              sails.log.error('A custom bodyParser can be configured without calling the wrapper function- e.g.:\n'+
-                '```\n'+
-                'bodyParser: require("connect-busboy")\n'+
-                '```'
-              );
-              sails.log.error(
-              'Alternatively, if you need to provide options:\n'+
-              '```\n'+
-              'bodyParser: {\n'+
-              '  fn: require("connect-busboy"),\n'+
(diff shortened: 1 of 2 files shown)


mikermcneil:
> But at this point our http hook is solid enough that it's safe to have middleware.bodyParser be the same as all the other middleware config: you just give it a function with req, res and next.

Agreed.  The documented behavior of custom middleware now is that it expects a req,res,next function directly (https://github.com/balderdashy/sails-docs/blob/master/concepts/Middleware/Middleware.md) .  So I think  the right approach is to change the documented example.  Skipper and the middleware config in Sails have bounced back and forth over time as our approach evolved, and this is definitely where we've always wanted to end up.

PR #3615, opened 2016-02-29
Ensure req._sails is present when responses hook is used

lennym:
The default responses use req._sails to provide access to the logger in order to write error cases to logs. Ensure that the req._sails property is present on the request object without creating a dependency on the request hook to be installed.


sgress454:
Merged #3599 instead

PR #3617, opened 2016-02-29
[Fixes #3604] Fix nested views

### LINKED ISSUE #3604: `res.view()` fails to render views nested more than 2 folders deep
See http://stackoverflow.com/questions/35637450/routes-in-sails-js-0-12-1

If you have a view in say `views/app/user/homepage.ejs`, and a route in `config/routes.js` like:

```
'/user': {view: 'app/user/homepage'}
```

and try to hit `/user` in a browser, Sails will attempt to display `app/user` (and fail because it doesn't exist) instead of displaying `app/user/homepage.ejs`.

The issue is in [this code](https://github.com/balderdashy/sails/blob/master/lib/hooks/views/onRoute.js#L55-L64) which only parses view expressions two levels deep.  However, because of a previous bug in 0.11.x that was [patched in 0.12](https://github.com/balderdashy/sails/commit/166e77255dc2860f92bc5cce23dc3878a1e5627f#diff-9dbb3a055ec218ae304a9ab6634afb34), this behavior actually worked in 0.11 (the bug prevented req.options from overriding options that were initially bound with the handler, so [this code](https://github.com/balderdashy/sails/blob/master/lib/hooks/views/actions.js#L63-L70) had no effect and the correct view path got sent to `res.view()`).

We can solve this by using Lodash `_.get()` to parse the view expression instead of just taking the top-level and second-level view names.

sgress454:
Now that `req.view` determines what to render based on what's in `req.options`, I don't see the utility of having a middleware function bound for each individual view, hence the removal of most of the code in `lib/hooks/views/action.js`.  If I'm missing something, lemme know.  Tests all pass without it though.  

Also added new tests for using the view target syntax (e.g. `'/foo': {view: 'foo/bar/baz'}`) and a test for the ability to do `'/foo': {view: 'foo'}` and have it render `views/foo/index.js` if it exists--not sure if that was ever documented, but the code was is in there so I made sure it still works.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### Master
 
+* [BUGFIX] Allow view target syntax for routes to specify deeply-nested views [3604](https://github.com/balderdashy/sails/issues/3604)
 * [BUGFIX] Allow custom bodyParser middleware config [3592](https://github.com/balderdashy/sails/issues/3592)
 
 ### 0.12.1
--- a/lib/hooks/views/actions.js
+++ b/lib/hooks/views/actions.js
@@ -9,68 +9,11 @@ var _ = require('lodash');
  */
 module.exports = function detectAndPrepareViews (sails, hook) {
   sails.modules.statViews(function (err, detectedViews) {
-    if (err) throw err;
+    if (err) {throw err;}
 
     // Save existence tree in `sails.views` for consumption later
     sails.views = detectedViews || {};
 
-    // Generate view-serving middleware and stow it in `hook.middleware`
-    createMiddleware(detectedViews);
   });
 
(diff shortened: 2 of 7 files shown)


mikermcneil:
@sgress454 looks great :fireworks: 

Re `index.ejs` routing: glad to have it tested.  Probably makes sense to make a note in the relevant route target documentation about this behavior as well- I'll dig that up real quick

mikermcneil:
@sgress454 here's [the change to sails-docs](https://github.com/balderdashy/sails-docs/commit/3bb7a76055a6731478a754732ef47d7c9de07dd1).

sgress454:
:egg:zellent

PR #3620, opened 2016-02-29
Proposal of cookie session paths

FranklinWaller:
Allow cookie sessions to be configured to a different cookie path.


sgress454:
Hi @FranklinWaller.  Mind putting a bit more description in the comments here?  See point 4 [here](https://github.com/balderdashy/sails/blob/master/CONTRIBUTING.md#pr-instructions) -- you can mostly re-use what you put in the backlog.  Thanks!

FranklinWaller:
The reason we need this is because we got a sailsjs oAuth2 ish server running and allows users to login. The whole application rests in /home. We also got an api tester which rests in /apitester with this you can test the api's that are used in /home. Now the session cookie interferes with this because it accepts requests because the session cookie is set (The api and the main application uses the same API). This way the /apitester is not reliable because it technically is already logged in. By setting the session cookie path to /home we can prevent /apitester from ever getting the session cookie and thus start from a complete blank. Now the apitester is one of the examples. But this could also count for "create a website" websites that allow javascript where the site doesn't use subdomains. 

The way to configure this could be configured in sails.config.session. For example:

`
module.exports.session = {
  key: 'homeonlycookie',
  path: '/home'
}
`

Cookies accept a "path" parameter.

FranklinWaller:
@sgress454 PR?

sgress454:
@FranklinWaller have you tried adding a `path` property to `sails.config.session.cookie`?  

```
// In config/sessions.js

 cookie: {
   path: '/home'
 }
```

I think this works.  If you can confirm, then we just need to add docs.

FranklinWaller:
Yes this works. A doc update would be great

sgress454:
Added docs here https://github.com/balderdashy/sails-docs/commit/188fb5df85697d7c9281f2e20e14a2989920ace9 and added comments to session config file.  Thanks for bringing this to our attention @FranklinWaller!

PR #3635, opened 2016-03-03
[proposal] Allow to bypass the Controller suffix convention

eric-burel:
Hi,

This might sound like a detail, but I think that the "Controller" suffix convention should be removed, and any .js file in the api/controllers folder should be considered as a controller by Sails.
The current default behaviour with suffix is not consistent with the behaviour of api/models and api/policies folders that consider any files inside them as a model or a policy, whatever their name is.
 Then, it is also redundant with the fact that controllers are in a folder already named controllers.

This way, one could name its controllers core.controller.js for example. Files that are used by controllers, but are not controllers or models themselves could be put into a api/helpers folder for example.

So far, I can't tell how to change configuration to have Sails to work this way. If Sails default behaviour can't be changed (and I guess that changing this convention would cost some efforts), at least adding a controller.js config file could help.


irlnathan:
Howdy @eric-burel. Thank you for bringing this up! We've [made a commitment](https://github.com/balderdashy/sails/blob/master/CONTRIBUTING.md#v-proposing-features-and-enhancements) not to change established conventions or default settings of Sails via the proposal process.  That said, a new configuration option like the one you mentioned could work.  What do you have in mind?

eric-burel:
Yes of course I don't want to throw an ideological debate ^^
After investigating the code, I think it could be implemented simply by soft-coding the regexp that filters controller in the loadControllers function :
 `filter: new RegExp('(.+)Controller\\.(' + sails.config.moduleloader.sourceExt.join('|') + ')$'),`
line 315 of the lib/hooks/moduleloader/index.js file.
Instead of controller it might be something like `sails.config.controllers.suffix || 'Controller'`

mikermcneil:
@eric-burel I like the idea of this (for example, I like to federate controllers sometimes, with separate actions in separate files).  I think that the first pass would be to implement a version of moduleloader hook that does it.  That keeps initial time investment to a minimum and let's folks start using this today in the current version of sails (i.e. by installing a separate package and overriding the core moduleloader hook).

If you wouldn't mind creating a repo with a standalone version of the moduleloader hook with your changes added, then it'll be easy for folks to start taking advantage of this patch today (and then meanwhile, we can get a good sense of how to approach module loading in the future).

eric-burel:
Hi,
I might give try this next time I use Sails, but that could take some time ^^ Are there some unit tests for the module loader ?

PR #3642, opened 2016-03-07
[proposal] Call `res.serverError` if a controller function returns a rejected Promise

not-an-aardvark:
At the moment, if a controller function throws a synchronous error, `res.serverError` is called automatically. This is useful because it ensures that a response always gets sent back to the client, and it removes the need for boilerplate try-catch statements.

However, this functionality does not currently exist for asynchronous controller functions. As a result, it's possible for an async function to silently throw an error, without ever sending a response.

Here's an example of a controller file to demonstrate why this can be an issue:

```js
// SomeController.js

exports.function1 = function (req, res) {
  // This automatically calls res.serverError under the hood
  throw 'something bad happened.';
};

exports.function2 = async function (req, res) {
  /* At the moment, this does not automatically call res.serverError.
  Instead, the error gets swallowed and no response is sent.
  A workaround is to wrap the entire function in a try-catch, but that leads to a lot of boilerplate code.
  (This example uses ES7 async functions, but the issue applies to any code that runs asynchronously.) */
  throw 'something bad happened asynchronously.';
};
```
To resolve this inconsistency, I propose that **controller functions should be able to return a Promise. If that Promise ends up rejecting, `res.serverError` is automatically called with the rejection reason, as if an error was thrown synchronously.** If this proposal were implemented, `function1` and `function2` above would both result in a call to `res.serverError`.

Mocha allows a similar type of thing, e.g.

```js
it('does some asynchronous test', function () {
  // The test suite succeeds if somePromise fulfills, and fails if somePromise rejects
  return somePromise;
});
```

I think this behavior would get rid of boilerplate controller code and help prevent silent errors, with no significant downsides.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -32,6 +32,7 @@ The backlog consists of approved proposals for useful features which are not cur
 Feature                                          | Proposal                                                                              | Summary
  :---------------------------------------------- | :------------------------------------------------------------------------------------ | :----------------------------------------------------------------------------------------------------------
  Generate `test/` folder in new Sails apps       | [#2499](https://github.com/balderdashy/sails/pull/2499#issuecomment-171556544)        | Generate a generic setup for mocha tests in all new Sails apps.  Originally suggested by [@jedd-ahyoung](https://github.com/jedd-ahyoung).
+ Call `res.serverError` if a controller returns a rejected Promise       | [#3642](https://github.com/balderdashy/sails/pull/3642)        | If a controller function returns a rejected Promise without sending a response, automatically call `res.serverError` with the rejection reason.
 
 
 


not-an-aardvark:
Not to rush the process if this is still being reviewed, but would this proposal be accepted if I implemented it and made a PR? I'd like to try and implement this if the proposal is a good idea, but I'd rather not waste time implementing it if the proposal itself is rejected.

sgress454:
Hi @not-an-aardvark, thanks for following up.  The basic rule here is, "if it's going to be a significant amount of work, and you haven't already started, then don't write any code until the proposal is merged into the backlog."  My instincts tell me that the amount of changes to core here would be significant.  Am I wrong on that?  Can this be done additively, without major changes to existing functionality?

mikermcneil:
@not-an-aardvark This is an interesting feature that is a valid approach to writing actions, particularly for those that are promise-inclined.  And while it could be done in an additive way, I'm hesitant-- mainly because of my personal experience with the request/response lifecycle in other frameworks.  I remember getting pretty confused in Grails with when the response was actually going to be sent- and since Node is asynchronous, I think that relying on a return value in core would be potentially confusing for many users, and could result in unexpected behavior when promises are inadvertently returned.

That said, this could definitely be implemented in a separate package, as an override for the controllers hook (e.g. `sails-hook-controllers-promises`).  And as long as compatibility for standard usage was maintained, blueprints and other core features would continue to work.  If there's interest in working on this, I'll spend some time making an example.

not-an-aardvark:
Thanks for the replies. I see what you mean with regard to how mixing callbacks and returned Promises can get confusing, so I'll close this PR then.

---

For anyone viewing this in the future, here's a workaround I made that implements this functionality:

```js
var Promise = require('bluebird');

// (this function is the same for every controller/policy, so it could be used as a service)
var funcMapper = func => (req, res, next) => Promise.method(func)(req, res).asCallback(next);

module.exports = _.mapValues({
  function1 (req, res) {
    doSomeLogic();
  },
  async function2 (req, res) {
    doSomeMoreLogic();
  }
}, funcMapper);
```

gleam-ru:
@not-an-aardvark, @mikermcneil 
How about this workaround?
It is a hook:
```js
/**
 * Handles async function errors
 * if they apper in controller`s actions
 *
 */
module.exports = function(sails) {
  return {
    initialize(cb) {

      var controllers = sails.hooks.controllers;
      controllers.middleware = _.mapValues(controllers.middleware, controller => {
        return _.mapValues(controller, action => {
          return async function(req, res) {
            try {
              await action.apply(this, arguments);
            }
            catch (err) {
              return res.serverError(err);
            }
          };
        });
      });

      return cb();
    },
  };
};

```

PR #3645, opened 2016-03-09
[patch] Fix incorrect handling of Error objects in handleError() and add test…

tapuzzo-fsi:
Fix incorrect handling of Error objects in handleError() and add test to prevent regression of same. With this change the error response created by the default response hooks will include more information when an error is encountered while processing a request.  The body of the HTTP response will include a message and stack (if not in production.)


mikermcneil:
@tapuzzo-fsi awesome, thank you!

PR #3646, opened 2016-03-09
[fixes #3174] findOne filtering

### LINKED ISSUE #3174: findOne filtering
So the `req.options.where` and `req.options.values`, are a gift for restricting access to data by policy, for example, i can restrict **update**, **create** and **find** to objects that are created by the user. 

But **findOne** only looks for a primary key, therefore restriction by policy becomes much harder and custom blueprints/controllers must be created.

This problem is easily corrected by adding a `.where( actionUtil.parseCriteria(req) )` in the default blueprint of the findOne action.

yodaddyp:
Add where filtering to findOne Blueprints so findOne can be protected by a policy.

### DIFF
--- a/lib/hooks/blueprints/actions/findOne.js
+++ b/lib/hooks/blueprints/actions/findOne.js
@@ -23,7 +23,8 @@ module.exports = function findOneRecord (req, res) {
   var Model = actionUtil.parseModel(req);
   var pk = actionUtil.requirePk(req);
 
-  var query = Model.findOne(pk);
+  var query = Model.findOne(pk)
+      .where( actionUtil.parseCriteria(req) );
   query = actionUtil.populateRequest(query, req);
   query.exec(function found(err, matchingRecord) {
     if (err) return res.serverError(err);


mikermcneil:
Hey @yodaddyp, thanks for taking the time to send this in.  Since this would be an API change, it would need to go through the proposal process.

In this case, it seems like the **find** blueprint with `limit: 1` would be an easier workaround-- does that make sense?

yodaddyp:
Hi @mikermcneil, wouldn't **find** with `limit: 1` return an array, where as **findOne** returns an object?

irlnathan:
@yodaddyp that's correct, `find()` returns an array and `findOne` returns an object (a.k.a. dictionary) :)

yodaddyp:
Sorry I don't think I understand then. If it returns a different data type I am not sure how that is a workaround? My current workaround at the moment is to override the findOne blueprint as mentioned in https://github.com/balderdashy/sails/issues/3174

mikermcneil:
@yodaddyp correct- if you want to continue to use the blueprint API for this endpoint, then in your front-end code / API docs, you would need to change the expected result from the API to a single-item or zero-item array vs. either a dictionary or 404.

Overriding the blueprint action with your own custom action is probably the right way to go.  If you're familiar with Rails, you might think of blueprints like scaffolds.  Since they're simpler and API-driven, blueprints are a bit more flexible, but still, they are not designed to work for every use case.  In a large production application, you might end up overriding 90-100% of your blueprint actions with custom actions, and you will probably disable blueprint routes.  In a very small application, like a brochureware website with some minor interactive features, or a simple mobile app that needs to save high scores, you might end up somewhere in the middle.  Hope that helps!

yodaddyp:
@mikermcneil thanks

PR #3647, opened 2016-03-10
[proposoal] Add the ability to use --staging as a command line argument to sails lift

jodonnell-broadsoft:
This PR makes it so you can do a `sails lift --staging` instead of having to use NODE_ENV=staging node app.js.

You can then create a file in config/env/staging.js that is a copy of production.js that suits your staging environment specific settings.

This works great but I feel we could further improve this by adding a `tasks/register/buildStaging` and `tasks/register/staging.js` as well as generating a sample `config/env/staging.js` for all the n00bs out there like myself.

As discussed with @mikermcneil over email. Hopefully this is how you envisioned it Mike. If not, please help me make it better so I can help. :)

### DIFF
--- a/lib/app/configuration/load.js
+++ b/lib/app/configuration/load.js
@@ -70,7 +70,7 @@ module.exports = function(sails) {
             port: overrides.port || undefined,
 
             // `--prod` command-line argument
-            environment: overrides.prod ? 'production' : (overrides.dev ? 'development' : undefined)
+            environment: overrides.prod ? 'production' : (overrides.staging ? 'staging' : (overrides.dev ? 'development' : undefined))
 
           });
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### Master
 
+* [ENHANCEMENT] Added the ability to use --staging and have the application use config/env/staging.js
 * [BUGFIX] Add blacklist to "update" blueprint action so that it can be used with primary keys that are not "id" [3625](https://github.com/balderdashy/sails/issues/3625)
 * [ENHANCEMENT] Allow hooks to be turned off by setting their environment var to the string "false" [3618](https://github.com/balderdashy/sails/issues/3618)
 * [BUGFIX] Allow view target syntax for routes to specify deeply-nested views [3604](https://github.com/balderdashy/sails/issues/3604)
--- a/bin/sails-www.js
+++ b/bin/sails-www.js
@@ -50,6 +50,9 @@ module.exports = function() {
     if (sails.config.environment === 'production') {
       overrideGruntTask = 'buildProd';
     }
+    else if (sails.config.environment === 'staging') {
+      overrideGruntTask = 'buildStaging';
+    }
     else {
       overrideGruntTask = 'build';
     }
--- a/bin/sails-www.js
+++ b/bin/sails-www.js
@@ -50,9 +50,6 @@ module.exports = function() {
     if (sails.config.environment === 'production') {
       overrideGruntTask = 'buildProd';
     }
-    else if (sails.config.environment === 'staging') {
-      overrideGruntTask = 'buildStaging';
-    }
     else {
       overrideGruntTask = 'build';
     }
--- a/bin/sails.js
+++ b/bin/sails.js
@@ -52,6 +52,7 @@ program.usage('[command]');
 var cmd;
 cmd = program.command('lift');
 cmd.option('--prod');
+cmd.option('--staging');
 cmd.option('--port [port]');
 cmd.unknownOption = NOOP;
 cmd.description('');
--- a/lib/hooks/grunt/index.js
+++ b/lib/hooks/grunt/index.js
@@ -70,9 +70,6 @@ module.exports = function(sails) {
       else if (sails.config.environment === 'production') {
         gruntTaskListName = 'prod';
       }
-      else if (sails.config.environment === 'staging') {
-        gruntTaskListName = 'staging';
-      }
       else {
         gruntTaskListName = 'default';
       }
--- a/lib/hooks/grunt/index.js
+++ b/lib/hooks/grunt/index.js
@@ -70,6 +70,9 @@ module.exports = function(sails) {
       else if (sails.config.environment === 'production') {
         gruntTaskListName = 'prod';
       }
+      else if (sails.config.environment === 'staging') {
+        gruntTaskListName = 'staging';
+      }
       else {
         gruntTaskListName = 'default';
       }
--- a/test/integration/www.test.js
+++ b/test/integration/www.test.js
@@ -127,25 +127,5 @@ describe('Running sails www', function() {
 				}
 			});
 		});
-
-		it('--staging should execute grunt buildStaging', function(done) {
-
-			// Move into app directory
-			process.chdir(appName);
-
-			// Overrwrite session config file
-			// to set session adapter:null ( to prevent warning message from appearing on command line )
-			fs.writeFileSync('config/session.js', 'module.exports.session = { adapter: null }');
-
-			sailsServer = spawn('node', [sailsBin, 'www', '--staging']);
-
-			sailsServer.stdout.on('data', function(data) {
-				var dataString = data + '';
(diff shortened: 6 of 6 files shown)


lib/app/configuration/load.js
```diff
@@ -70,7 +70,7 @@ module.exports = function(sails) {
             port: overrides.port || undefined,
 
             // `--prod` command-line argument
-            environment: overrides.prod ? 'production' : (overrides.dev ? 'development' : undefined)
+            environment: overrides.prod ? 'production' : (overrides.staging ? 'staging' : (overrides.dev ? 'development' : undefined))
```
jodonnell-broadsoft: should there be a test added for this?

jodonnell-broadsoft:
Spelled proposal wrong :oncoming_police_car:

RWOverdijk:
lgtm :+1:

jodonnell-broadsoft:
thanks! @RWOverdijk :smile:

mikermcneil:
@RWOverdijk thanks :+1: 

@jodonnell-broadsoft looks good.  One relatively trivial thing that would be a good idea while you're in there is to change the comment in `lib/app/configuration/load.js` to something like:

```javascript
// --prod, --dev, and --staging CLI args
```


#### NODE_ENV vs. sails.config.environment

I still like what you and I discussed re: allowing `sails.config.environment === 'staging'` while `process.env.NODE_ENV === 'production'`  (e.g. lifting with `NODE_ENV=production sails_environment='whatever' sails lift`).  I think it'd be great to get that in as part of this PR (or separately, whatever makes sense).

The plan I originally sent you was a little incorrect, I just realized.  Here's a corrected version as a recap:

> 1. The code that automatically sets `sails.config.environment` to whatever NODE_ENV is _only does so if `sails.config.environment` was not explicitly set_.  This is good, and what we want (see https://github.com/balderdashy/sails/blob/master/lib/app/configuration/load.js#L132)
>
> 2. However, we should change the code a few lines after that that automatically sets `NODE_ENV` based on `sails.config.environment`.   Instead of **always** setting `NODE_ENV` to be the same as `sails.config.environment`, we should only do so `if ( !process.env.NODE_ENV )` (i.e. the NODE_ENV env var was not explicitly set). I believe the relevant change would need to occur in the if statement here: https://github.com/balderdashy/sails/blob/master/lib/app/configuration/load.js#L150   (i.e. change it to the prettified/properly-formatted version of this: `if (mergedConfig.environment && !process.env.NODE_ENV)`)


Then, to lift in staging, using the realistic NODE_ENV that will actually be used in production (i.e. which affects dependencies), you could do:
```bash
NODE_ENV=production sails lift --staging

# or:
# NODE_ENV=production sails_environment=staging sails lift

# or:
# put `"environment": "staging"` in your .sailsrc file
```


#### Re: the grunt stuff:

How about we change the Grunt stuff in `sails www` to match the [way it works in the Grunt hook](https://github.com/balderdashy/sails/blob/master/lib/hooks/grunt/README.md#run-a-grunt-tasklist-when-sails-is-loaded)?  (See also the relevant code [here](https://github.com/balderdashy/sails/blob/master/lib/hooks/grunt/index.js#L55-L75))

We obviously can't do _exactly that_ (i.e.  `default.js` / `prod.js` / `<<SAILS.CONFIG.ENVIRONMENT>>.js`), but we can do something comparable, e.g:

```javascript
      var pathForEnvSpecificTaskList = path.resolve( sails.config.appPath, path.join('tasks/register/','build'+_.capitalize(sails.config.environment) + '.js') );
      if ( fs.existsSync(pathForEnvSpecificTaskList) ) {
        gruntTaskListName = 'build'+_.capitalize(sails.config.environment);
      }
      // If the environment is "production", use the "prod" tasklist (unless a "production" tasklist exists)
      else if (sails.config.environment === 'production') {
        gruntTaskListName = 'buildProd';
      }
      else {
        gruntTaskListName = 'build';
      }
```

So then:

```bash
# run the tasklist at `tasks/register/build.js`:
sails www


# try to run the tasklist at `tasks/register/buildStaging.js`, fall back to `tasks/register/build.js`:
sails www --staging
NODE_ENV=production sails www --staging  # same as above


# try to run the tasklist at `tasks/register/buildProduction.js`, fall back to `tasks/register/buildProd.js`:
NODE_ENV=production sails www
sails www --prod    # same as above
NODE_ENV=production sails www --prod  # same as above
```


#### Generators
re: generators, I think given the grunt stuff I mentioned above, we don't actually need to change the generators at all (if we change them, I fear that this could eventually be one of those slippery slope kinds of things).


#### Docs
That said, we definitely need to update the relevant docs to reflect:
+ `sails www` now uses `sails.config.environment` to determine the appropriate build tasklist to run (but is backwards compatible with `buildProd.js` and `build.js`)
+ `sails.config.environment` can now be set separately from NODE_ENV (but NODE_ENV will still be automatically set if you didn't specify it)
+ the new `--staging` option for Sails lift

Here's all the places I can think of offhand that will need to be updated:
+ I spent some time thoroughly documenting this behavior when working on the [internal contributor docs for the Grunt hook](https://github.com/balderdashy/sails/blob/master/lib/hooks/grunt/README.md#run-a-grunt-tasklist-when-sails-is-loaded), and just added a reference to it [here](https://github.com/balderdashy/sails-generate-frontend/blob/master/templates/tasks/register/default.js#L6) in the `default.js` tasklist generated in new sails projects.  It would be great to do the same thing for [the `build.js` tasklist](https://github.com/balderdashy/sails-generate-frontend/blob/master/templates/tasks/register/build.js#L12) as well as the [`buildProd.js` tasklist](https://github.com/balderdashy/sails-generate-frontend/blob/master/templates/tasks/register/buildProd.js#L6).
  + The examples should show `sails.config.environment` rather than `NODE_ENV` 
+ apparently we don't have a reference page for `sails www` in http://sailsjs.org/documentation/reference/command-line-interface, so it'd be great to add one and include the relevant information.
+ we should mention this in [concepts/assets/default-tasks](http://sailsjs.org/documentation/concepts/assets/default-tasks)
+ I think we need to check the anatomy docs, the [concepts/configuration]() docs, and also anywhere else I'm forgetting :p


#### Tests
We should add a test or two to `test/integration/lift.lower.test.js`.  When I added the behavior that restores NODE_ENV if it was automatically set, I added [this one](https://github.com/balderdashy/sails/commit/357b30da27dceb2d1820499a35988ab2dca5873b#diff-fce1671730063f26f8a812065fc9257aR67).  But now we should also ensure that NODE_ENV _is not automatically set **at all**_ if it was explicitly specified;  i.e. that you can load and/or lift an app with `environment: 'staging'` and `NODE_ENV=production`, and NODE_ENV will remain untouched.



@jodonnell-broadsoft thanks for all your help!  Looking forward to this, it's awesome.

jodonnell-broadsoft:
@mikermcneil Thanks for the feedback on this. I will look into this presently and give you a V2 of this Pr. Glad this is getting into the code. Should be cool :+1:

joncodo:
@mikermcneil 

First of all, thanks so much for making my open source experience awesome. I really appreciate the time you are putting in with me :smile: . I know you could do this yourself much faster.

I made the changes you suggested and I think I am on the right track. Now need to add the docs and the tests. Will hopefully do that tomorrow. Also pulled in latest master if you want to review.

Thanks again guys :)

joncodo:
Seems like merging has changed some stuff that I did not intend. Will revert the merge and try again tomorrow.

joncodo:
I see now that master is also failing. Added this small PR to add the travis build flag to the readme page. #3652

joncodo:
And here is a docs PR. https://github.com/balderdashy/sails-docs/pull/665

joncodo:
So I have a failing test.

should revert NODE_ENV env variable if it was set automatically on load/lift

in lift.lower.test.js

This leads me to believe I have missed the case somewhere that the process.env.NODE_ENV should be reverted. Not 100% sure on the fix here. Would one of you guys mind taking a look? @RWOverdijk or @mikermcneil ?

Other than that, I think that all of your concerns should be resolved. Also added a docs PR that is linked here. :)

joncodo:
Will pull in latest on this and let you know.

albertpeiro:
+1 for this issue. What's the state of this?

PR #3648, opened 2016-03-10
[proposal] Force async-ness in template rendering methods

lennym:
There is a long-standing open bug in consolidate - https://github.com/tj/consolidate.js/issues/143 - which causes errors in the callback functions passed to rendering methods to propogate back into the try/catch contained in consolidate and cause the callback to be fired twice. Since this bug has been open in consolidate for over two years, not much hope should be held of having it resolved there.

As such, add protection in Sails against errors in the callback functions being caught inside consolidate and the callback fired twice by forcing the callback to be called asynchronously.

Example:

```javascript
res.view('template', {}, function (e, result) {
  if (e) {
    // error thrown below causes callback to be fired twice and error caught here
  }
  throw new Error('error');
});
```

This can cause errors in userland code to be incorrectly attributed to template rendering and makes debugging difficult.

I appreciate that this *should* be fixed at the root in consolidate, but after 2 years, and with the [maintainer insisting it is not a bug](https://github.com/tj/consolidate.js/issues/143#issuecomment-37481417), I don't hold much hope for that.

### DIFF
--- a/lib/hooks/views/configure.js
+++ b/lib/hooks/views/configure.js
@@ -73,7 +73,10 @@ module.exports = function configure ( sails ) {
     }
 
     // Save reference to view rendering function
-    sails.config.views.engine.fn = fn;
+    // munge callback to protect against https://github.com/tj/consolidate.js/issues/143
+    sails.config.views.engine.fn = function (str, options, cb) {
+      fn(str, options, function (e, result) { setImmediate(function () { cb(e, result); }); });
+    }
     sails.log.silly('Configured view engine, `' + engineName + '`');
   }
 


mikermcneil:
@lennym I've definitely noticed some very strange behavior with this before (as you can probably tell from the shifting sands in the implementation of res.view()).  Finally, we just removed official support for the callback argument to `res.view()`.  But I'd like to have this work (esp because then we could properly support e.g. `sails.compileView('stuff/whatever', {}, function (err, compiledHtml){  /* ... */ })` for use with transactional emails)

That said, I don't want to introduce anything that could crash the process...  To clarify, is the issue that:

+ (A) the third arg [here](https://github.com/lennym/sails/commit/cd413e15435947aa855e27aab16d9cd9e65ad493#diff-e210235c7225973db791ddaf6a65b40eR77) (`cb`) is throwy?
+ Or is it (B) that the userland function provided as the third argument to `res.view()` / `res.render()` might be throwy, and if it throws it ends up calling itself again?

If it's (B) the latter case (which I think it is, based on your detailed explanation, and also because I'm having some deja vu of breaking bottles and screaming about it), then absolutely, let's get this merged in asap.  If it's (A), the former case, I think we need to figure something else out-- unless (A) and (B) are actually the same thing (i.e. the function is provided from userland as the 3rd arg to `res.view()` or `res.render()` in both cases), in which case we're also good to merge this.

In other words, I'm 100% down unless making this change allows view engines to throw uncaught errors where we weren't before.  Which I don't think it does.  But.. yeah, just checking.

lennym:
Thanks @mikermcneil 

Firstly, I would very strongly advocate for preserving methods for detaching template rendering from the response. Such things are fantastically useful in a lot of scenarios. Our current use case is actually a BDD test scenario editing tool where we construct fully formed feature files by building templates out of fragments of user input. The issue became apparent when we were trying to validate the built result with a parser that threw for invalid gherkin syntax and caused all manner of odd behaviour.

Anyway, the issue is very definitely the latter (B). There is a contiguous, synchronous call stack from the user-provided callback (or even sails' callback - this would manifest bugs in sails' implementation of res.view as engine errors) back into the try/catch inside consolidate. It will not allow errors throw inside the engines to be uncaught - this would be thrown before the forced async, and so would be passed as a first arg to the callback as at present.

Thanks again.

mikermcneil:
> I would very strongly advocate for preserving methods for detaching template rendering from the response

Completely with you- I just want to expose a cleaner API that doesn't rely on access to `res`.

> Anyway, the issue is very definitely the latter (B)

Sweet, just checking :+1:

mikermcneil:
@lennym on a separate note, we added [this](https://github.com/balderdashy/sails/blob/master/lib/hooks/views/index.js#L78) as an experimental feature a while back.  The issue, as you can see [here](https://github.com/balderdashy/sails/blob/master/lib/hooks/views/render.js#L9) is that the logic is duplicative.  The stuff about mocking the request and everything is totally fine, I think we should keep that.  Really the only thing that's kept me from cleaning that up and making it official is that I kept running up against problems/complexities with the various intertwined dependencies (e.g. like the one you just addressed.)

As for `sails.renderView()` itself, the API needs to be cleaned up (ignore most of the comments in there).  What I would suggest is something like this:

```javascript
// "compile" instead of "render" to avoid confusing folks new to Sails who might think
// "render" involves doing something with the response
sails.compileView('foo/bar', {
  whateverLocals: [{you: 'want', go: 'here'}]
}, function afterwards(err, html){
  if (err) {
    // handle
    return;
  }

  // e.g. now send HTML email
});
```

PR #3652, opened 2016-03-14
[proposal] Adding the travis master build flag to the README for master branch.

joncodo:
It seems that this does not exist and I was tricked into thinking my branch was broken. It seems that pulling master into my branch broke it from what I can tell. This may be a great thing to have on our homepage. Keeps us honest.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 [![Gitter](https://badges.gitter.im/Join Chat.svg)](https://gitter.im/balderdashy/sails?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
 [![NPM version](https://badge.fury.io/js/sails.svg)](http://badge.fury.io/js/sails)
-
+![Travis](https://travis-ci.org/balderdashy/sails.svg?branch=master)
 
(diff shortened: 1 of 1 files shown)


sgress454:
Hi @joncodo.  The Travis build status for Sails and all of its components is shown on the [Modules](https://github.com/balderdashy/sails/blob/master/MODULES.md) page.  I'd be fine moving showing it in the README as well if we hadn't been having so many problems with Travis lately -- it just seems to fail randomly in ways that are impossible to replicate locally, even using their Docker image.  At the point where we can confidently say that the Travis badge 100% reflects the status of Sails, we'll revisit this.

mikermcneil:
@joncodo 👍  been working on this, so far still searching.  We spent a lot of time earlier this year cleaning up tests and I think we're close to having things on Travis be accurate-- but a lot of the problem is that a lot of our tests are rather old (and of course, it feels horrible to delete a working test).  So what I think needs to happen is gradual replacement of the oldest Sails tests.  Once we've completely eliminated the race conditions, I'm 100% ok with putting the build light on the README (I'd like to keep it on MODULES.md as well though)

Issue #2576, opened 2015-01-20
routesDisabled does not property bypass CSRF


adamcoomes:
Using the "routesDisabled" property in the CSRF config doesn't seem to work in testing. I'm guessing it's some sort of bug.

sgress454:
Hm, it [passes our test](https://github.com/balderdashy/sails/blob/master/test/integration/hook.cors_csrf.test.js#L774)...what's your CSRF config look like?

adamcoomes:
module.exports.csrf = {
    protectionEnabled: 'true',
    routesDisabled: '/api/link/imageCallback'
}

Essentially imageCallback is an action in my LinkController that is called from an external source via a POST. However even with this config, i'll get a CSRF Mismatch when this route is hit by the source.

sgress454:
Works For Me™.  Check that your route configuration is correct, and that your external source is hitting the right URL?  The CSRF Mismatch response will show up in place of 404 when CSRF is on, so you might not realize you're hitting the wrong page (or have the route misconfigured).

adamcoomes:
Ah, knowing that is interesting. Maybe it's not a CSRF Mismatch but everything works with CSRF off so that's what makes me think it's a CSRF issue.

fundon:
`routesDisabled` just matches realpath (e.g. /users, /upload)

`routesDisabled` does not match  `/users/:id`.

@sgress454  Please improve this.

bvallelunga:
+1

bvallelunga:
@sgress454 is there a way to set a prefix like "/api" where csrf will be disabled?

ivanjh:
@bvallelunga You could disable the built-in csrf hook and create a user hook (copy/paste/hack) that does the same job, but performs a prefix match on routesDisabled instead of full match.

bvallelunga:
this is what I have working currently

``` js
   csrfChecker: function(req, res, next) {
      var routesDisabled = sails.config.csrf.routesDisabled

      if(req.path.indexOf("/api") > -1 && routesDisabled.indexOf(req.path) == -1) {
        if(routesDisabled == "-")
          routesDisabled = req.path
        else
          routesDisabled += ("," + req.path)

        sails.config.csrf.routesDisabled = routesDisabled
      }

      next()
    },
```

JerryC8080:
routesDisabled does not match /users/:id. 
+1

sailsbot:
[...]

PR #3661, opened 2016-03-16
[implements #2576] sails.config.csrf.routesDisabled support RegExp

bolasblack:
Make `sails.config.csrf.routesDisabled` support RegExp, a partial solution of #2576 , because I think maybe the format of `/:id` is useless...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### Master
 
+* [ENHANCEMENT] Allow use Express style path and RegExp in `sails.config.csrf.routesDisabled` [d3f634c](https://github.com/balderdashy/sails/commit/d3f634c9ac0c5e2172710fe27ab3f61f8303d840)
 * [BUGFIX] Fix typo which could cause crashing when attempting to serialize non-json-compatible output in the response to a socket request.
 * [UPGRADE] Upgraded version of Consolidate to `0.14.1` [a70623c](https://github.com/balderdashy/sails/commit/a70623ce2809d497b3581268354f06904d862268)
 * [UPGRADE] Upgraded version of grunt-contrib-watch to `1.0.0` [3678](https://github.com/balderdashy/sails/issues/3678)
--- a/lib/hooks/csrf/index.js
+++ b/lib/hooks/csrf/index.js
@@ -5,7 +5,8 @@ module.exports = function(sails) {
    */
 
   var _ = require('lodash'),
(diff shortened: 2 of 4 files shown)


sgress454:
Hi @bolasblack -- thanks for the PR.  I think any solution here would have to use [path-to-regex](https://github.com/pillarjs/path-to-regexp) so that actual Express-style dynamic URLS could be matched.  That's what folks have been missing.

bolasblack:
@sgress454 WOW, awesome, this package make things easy, I will change my PR to use this package.

Thanks again!

mikermcneil:
@bolasblack thanks for following up!  This looks great.  And `path-to-regex` is indeed a lifesaver. 

Would you mind rebasing this PR with the latest on master?  Then if you or someone else reading this has a sec to update the [csrf config reference docs](https://github.com/balderdashy/sails-docs/blob/master/reference/sails.config/sails.config.csrf.md), this is good to go 👍  for the next release.

bolasblack:
@mikermcneil 👌 now

sgress454:
Looking forward to getting this in as well -- I am wondering though if we couldn't pull some of the logic out of the wildcard route handler and have it just run once at the top of the initialize.  Specifically it seems sub-optimal to compute the `disabledRoutes` array on every request, and to run `pathToRegexp` on all of the string values in that array on every request.  The disabled routes configuration isn't going to change between requests, and even if it did, `disabledRoutes` would ideally just be reloaded as needed, not every time.

bolasblack:
@sgress454 You are right 👍

brettimus:
It looks like someone [added a print statement](https://github.com/balderdashy/sails/blob/master/test/hooks/views/skipAssets.test.js#L71) to check response headers just above the line where this branch previously crapped out. Any chance the travis build may magically pass if re-run?

bolasblack:
@brettimus 🍻  test passed now

brettimus:
🙌🙌🙌

brettimus:
How does merging happen? ((Asking for a friend 😁 ))

bolasblack:
@brettimus I'm waiting too 😂

sgress454:
Looks good!  Updated [master docs](https://github.com/balderdashy/sails-docs/blob/master/reference/sails.config/sails.config.csrf.md).  Thanks for your patience!

bolasblack:
🙌🙌🙌🙌🙌🙌

PR #3679, opened 2016-04-01
[patch] Sails.prototype.inspect() - fix link to docs

wulfsolter:
Sails.prototype.inspect() which is triggered by running something like `console.log(sails);` links to old docs page (http://links.sailsjs.org/docs) from 2 years ago which is now non-existant. This patch changes link to http://sailsjs.org/documentation/concepts/


wulfsolter:
ok, fixed!

mikermcneil:
@wulfsolter ahh yes, good ole "lighthouse"-- not the smartest thing I ever did.  Thanks!

PR #3681, opened 2016-04-01
[Patch] Remove needless require from mocha.opts

markelog:


mikermcneil:
@markelog looks good, thank you!

PR #3720, opened 2016-05-02
[fixes #3590] Waits for Redis client connection before lifting when using connect-redis for session

### LINKED ISSUE #3590: No error shown upon `sails lift` when connect-redis is unable to connect to Redis
See #3589

When session adapter is set to `redis` in `config/sessions.js` and `connect-redis` is properly installed, but cannot connect to a Redis instance using the configured settings (perhaps because no Redis instance is running!) no error is displayed in the console, making it appear as if the server lifted without incident when in fact the app will be unable to handle sessions.

sgress454:
When using Redis adapter, inspects the `client` property and adds connection and error handlers.  Future-proofed against changes to connect-redis, and backwards compatible with cases where the user supplies an existing Redis connection in the session config as `sails.config.session.client`.

Would be nice to have tests for this, maybe using a Docker container...


sgress454:
refs #3590

lib/hooks/session/index.js
```diff
@@ -187,6 +187,29 @@ module.exports = function(app) {
       // Expose hook as `sails.session`
       app.session = SessionHook;
 
+      // If using Redis, wait for a connection or error
+      if (app.config.session.adapter === 'connect-redis' &&
+          // Future-proof by ensuring that there is a client object, and that it is an event emitter
```
mikermcneil: @sgress454 lgtm, just don't forget to use `app` (to avoid relying on the sails global)

lib/hooks/session/index.js
```diff
@@ -187,6 +187,29 @@ module.exports = function(app) {
...
+        return (function() {
+          // If the client is already connected (because it was provided in the config)
+          // then just continue
+          if (sails.config.session.store.client.connected) {return cb();}
+          // Define an error handler for the Redis client which will stop Sails from lifting
+          // by returning an error in the hook callback
+          var redisErrorHandler = function(err) {
+            return cb(err);
+          };
+          sails.config.session.store.client.on('error', redisErrorHandler);
```
mikermcneil: Actually, just noticed: we'll have to put in a spinlock here as well.  As a basis, here's an implementation of this I recently put together: https://github.com/mikermcneil/machinepack-redis/blob/f4dc7087df27694d5c723784e99a116213d4910e/machines/get-connection.js#L98 

(^^ I hand-tested that code against redis crashes/shutdowns that occur during and after connection, so should be good to go)

lib/hooks/session/index.js
sgress454: Got it, thanks.

lib/hooks/session/index.js
```diff
@@ -190,21 +190,21 @@ module.exports = function(app) {
...
+          app.config.session.store.client &&
+          app.config.session.store.client instanceof require('events').EventEmitter) {
         return (function() {
           // If the client is already connected (because it was provided in the config)
           // then just continue
-          if (sails.config.session.store.client.connected) {return cb();}
+          if (app.config.session.store.client.connected) {return cb();}
           // Define an error handler for the Redis client which will stop Sails from lifting
           // by returning an error in the hook callback
           var redisErrorHandler = function(err) {
```
mikermcneil: Actually, just noticed: we'll have to put in a spinlock here as well. As a basis, here's an implementation of this I recently put together: https://github.com/mikermcneil/machinepack-redis/blob/f4dc7087df27694d5c723784e99a116213d4910e/machines/get-connection.js#L98

(^^ I hand-tested that code against redis crashes/shutdowns that occur during and after connection, so should be good to go)

sgress454: It removes the error handler upon successful connection; do you think we need to remove the connection handler on error?

mikermcneil: @sgress454 Couldn't say off the top of my head.  I tested the code in mp-redis pretty thoroughly though, killing redis server mid-connection and post-connection, so I'm pretty confident what's in there does the trick as far as that.  I think the answer is just to try killing the redis server mid-lift and also after lift and just see what it does.  Maybe @luislobo has more feedback there?

lib/hooks/session/index.js
```diff
@@ -187,6 +187,29 @@ module.exports = function(app) {
...
+          sails.config.session.store.client &&
+          sails.config.session.store.client instanceof require('events').EventEmitter) {
+        return (function() {
+          // If the client is already connected (because it was provided in the config)
+          // then just continue
+          if (sails.config.session.store.client.connected) {return cb();}
+          // Define an error handler for the Redis client which will stop Sails from lifting
+          // by returning an error in the hook callback
+          var redisErrorHandler = function(err) {
+            return cb(err);
```
mikermcneil: Just double-checked this and I'm pretty sure we need to also handle `.on('end',`, and a few other crazy edge cases; pretty much everything in between line 98 and 158 here: https://github.com/mikermcneil/machinepack-redis/blob/f4dc7087df27694d5c723784e99a116213d4910e/machines/get-connection.js#L98-L158

PR #3730, opened 2016-05-12
[patch] Hooks README: remove duplicate sentences

straydogstudio:
See title.

### DIFF
--- a/lib/hooks/README.md
+++ b/lib/hooks/README.md
@@ -7,14 +7,10 @@
 
 ## Purpose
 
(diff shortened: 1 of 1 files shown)


sgress454:
Hi @straydogstudio, perhaps it's a bit confusing, but the sentences aren't duplicates.  The first is saying that most of the core features of Sails have been pulled into hooks.  The second is saying that those hooks (which currently still live in the core repo) may eventually be pulled into completely separate modules, like `sails-hook-sockets` and `sails-hook-orm`.

PR #3734, opened 2016-05-17
[proposal] Support PATCH for REST blueprints

jelhan:
Adds support for PATCH HTTP method to REST blueprints.

E.g. JSON API spec requires a PATCH request to update a resource:

> A resource can be updated by sending a PATCH request to the URL that represents the resource.
> http://jsonapi.org/format/#crud-updating


RWOverdijk:
If this gets in, PUT must be changed to overwrite a resource. Otherwise we'll be implementing half the spec.

sgress454:
Hi @jelhan, we ended up implementing this in 1.0 after the blueprint code had changed a bunch, but I wanted to make sure and merge your branch in since it was your idea in the first place ;)

We've kept support for update via `PUT` for Sails 1.0 to make it backwards compatible, but you'll get a deprecation warning if you access the blueprint via PUT.

PR #3758, opened 2016-06-08
[patch] Fix err variable using parent scope because of mispelled parameter

dustinws:
Hello!
A coworker was getting some weird errors when he tried to lift a second app on a port that was already in use. It was odd to me that this error didn't propagate to the `err` parameter in the sails lift callback. After checking out the source, it appears that there was a typo in the final handler of `sails.runBootstrap`. Instead of receiving the `err` value from the async call on line 82, the `err` variable is instead referencing the one defined by `runBootstrap` on line 68.
I hope this helps!

### DIFF
--- a/lib/app/private/initialize.js
+++ b/lib/app/private/initialize.js
@@ -84,7 +84,7 @@ module.exports = function initialize(cb) {
         return next();
       }
       return sails.hooks[hookName].handleLift(next);
-    }, function (er){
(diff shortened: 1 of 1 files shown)


mikermcneil:
@dustinws _very_ good catch.  Thanks so much!

dustinws:
Glad to help, thanks!

PR #3764, opened 2016-06-14
[patch] Use res.forbidden() instead of res.send(403)

wulfsolter:
Default behaviour of `policy === false || policy === null` is to res.send(403).
This bypasses responses/forbidden.js and it's associated view.


mikermcneil:
@wulfsolter Looks good, and makes sense.  Only thing to watch out for is that this could cause problems when the `policies` hook is enabled but the `responses` hook is not.  Probably as simple as a `if (_.isFunction(res.forbidden)) {...}` check, so I'll go ahead and merge and then check it out.  Thanks!

PR #3770, opened 2016-06-21
[proposal] remove wrench from test runner requirements

Ignigena:
Since `wrench` has been [officially deprecated](https://www.npmjs.com/package/wrench#warning-this-project-is-deprecated), I have removed it in favor of `fs-extras` which is already installed in the development dependencies.

I noticed there were several `TODO` comments mentioning that this should be done, and so here is a PR to do just that!

### DIFF
--- a/package.json
+++ b/package.json
@@ -112,8 +112,7 @@
     "root-require": "0.3.1",
     "should": "5.2.0",
     "socket.io-client": "1.4.6",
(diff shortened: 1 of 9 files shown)


sgress454:
Ran this with `wrench` completely removed and all tests passed, so LGTM.  Merging.  Thanks @Ignigena !

PR #3785, opened 2016-07-09
[proposal] Compression middleware enabled in production by default

awer91:
[proposal] Enable compression middleware in production env.
``compress: IS_NODE_ENV_PRODUCTION && require('compression')(),``
Expression ``var result = someFn && bool`` always will be `bool`,  current Sails compression middleware always stay disabled. 
My proposal to replace current Sails compression middleware by new working one. 

[proposal] Also i added some configuration for compression middleware. (optional)
sails.config.http.compress.compressionLevel - level of compression, defaults to 6,
sails.config.http.compress.chunkSize - min compressible filesize,
sails.config.http.compress.filter - filter fn

### DIFF
--- a/lib/hooks/http/get-configured-http-middleware-fns.js
+++ b/lib/hooks/http/get-configured-http-middleware-fns.js
@@ -163,7 +163,7 @@ module.exports = function getBuiltInHttpMiddleware (expressRouterMiddleware, sai
      /** Gzip response body, if  response has  "compressible" type **/
     compress: function(req, res, next){
        //skip this middleware in DEV mode
-       IS_NODE_ENV_PRODUCTION  return next();
+       if(IS_NODE_ENV_PRODUCTION === false)  return next();
        // include compression mod
        var compressor = require('compression');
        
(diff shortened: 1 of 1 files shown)


awer91:
fixed

awer91:
ok, fixed

awer91:
ok, fixed!

sgress454:
Hi @awer91, thanks for posting!

The expression `IS_NODE_ENV_PRODUCTION &&  require('compression')()` returns the result of `require('compression')()` (a middleware function) if `IS_NODE_ENV_PRODUCTION` is `true` (see the [compression middleware tests](https://github.com/balderdashy/sails/blob/master/test/integration/middleware.compression.test.js)).

Re: the extra config -- I definitely see the benefit of being able to fine-tune your compression.  We're making an effort to keep Sails light and modular, so if the defaults for some particular middleware don't work, we'd rather you just override that middleware entirely (see the [`bodyParser` config example](http://sailsjs.org/documentation/anatomy/my-app/config/http-js)).

PR #3786, opened 2016-07-09
[proposal]  Change compression middleware order

awer91:
[proposal]  Change compression middleware order
On default order, response files will not compress if "Cache-Control" header present, but conventionaly "compression" middleware should be loaded before  "static" or "router" middlewares.
This update helps to compress assets and use browser caching for assets same time.


jrettenmyer22:
I

awer91:
I was tested this. 

```javascript
//caching will work, compression will not
app.use(express.static(__dirname + '/public', { maxAge: 84000 }));
app.use(express.compress()); 
```

```javascript
//caching will work with compression
app.use(express.compress());  
app.use(express.static(__dirname + '/public', { maxAge: 84000 }));
```

PR #3793, opened 2016-07-20
[proposal] Typescript support

dsci:
Hi. 

I added TypeScript support by using the `typescript-require` module. Maybe there are some better solutions available or possible, it solves my needs. It's possible now to use *.ts files. 

Is this something that you may find interesting to merge?

### DIFF
--- a/lib/hooks/moduleloader/index.js
+++ b/lib/hooks/moduleloader/index.js
@@ -100,6 +100,11 @@ module.exports = function(sails) {
           extensions: ['ls'],
           module: 'LiveScript',
           require: 'livescript'
+        },
+        {
+          extensions: ['ts'],
+          module: 'typescript-require',
+          require: 'typescript-require'
         }
       ];
 


radarsu:
I was about to create same pull. Please introduce typescript support in next version of sails.

asemoon:
Sails.js has the potential to be the defacto web framework if they consider adding full support for typescript. it makes the technology future friendly and a lot of developers who are running their front-end on angular 2 would be start using sails.js.

mikermcneil:
@dsci @asemoon @radarsu Hey guys, thanks for sharing your thoughts.  The plan right now for Sails v1 is to continue to support typescript, coffeescript, and any other language via `app.js`.  I think the main issue here is documentation, both on the website and in the form of ready-to-uncomment code in app.js.  Please let me know if any of that seems off-base or if I'm missing something.  More info here: https://github.com/balderdashy/sails/pull/3142#issuecomment-247690755

radarsu:
I would like just to inform that my sails typescript project is not using `app.js`, but instead - `app.ts` file.

It works better, because there is no need to compile files remotely after every change, so I do not have "src" and "dist" destinations in my project. Just src is enough.

Compilation is done in background by `ts-node`, what also allows me to launch sails built with only `.ts` files with `pm2`.

mikermcneil:
@radarsu thanks for the correction-- right, in Sails v1, you'll be able to use whatever entry point you like.  As long as your `app.js`/`app.ts`/`app.coffee` file includes `require('typescript-require');` before doing a programmatic lift, then it'll work as expected expected.  That way, the moduleloader will no longer care what file extension you're trying to load, and Sails apps will be free to use any mix of files, whether that's JavaScript, JSON, TypeScript, ES6/ES7 via Babel or the transpiler of your choice, CoffeeScript, Literate CoffeeScript, Iced CoffeeScript, Iced Literate CoffeeScript, LiveScript, etc.

pulasthibandara:
Hi can you provide some pointer on how to implement sails on TS? I've been reading the docs and the web for few days but couldn't find anything :)

radarsu:
For me best way after long research was to inject into node_modules/sails/lib/hooks/moduleloader.js
line 180: conf.configExt = [(added)'ts', 'js','json'].concat(detectedExtens);
line 181: conf.sourceExt = [(added)'ts', 'js'].concat(detectedExtens);

Then, you can launch project with `ts-node app.ts` or even `pm2` with typescript module installed.

Only thing you should do is change all scripts in /config, /api, etc. to .ts format.

When I was testing few configurations, I made this: https://github.com/radarsu/sails-typescript-starter
It is not the best configuration (that I'm currently using), but it was quite good starting point.

You can also check this out: https://github.com/radarsu/dunthag-server (more recent).

Verdier:
The worst point with this approach is debugging. We use WebStorm and haven't find any solution for now...

There is an other working solution: changing paths in .sailsrc and run a grunt task before the lift process to compile `ts` to `js` with source map generation, then WebStorm is smart enough to add breakpoints directly in `ts` files.

pulasthibandara:
Thanks guys. I'll surely look into this method. While converting to JS method does not mess with the source, I'm more inclined to use a purely TS solution. But I'll explore both ways before comiting to the project.

sgress454:
Hi @dsci, the handling for different JavaScript dialects has changed in 1.0.  Basically, Sails will load _any_ file in your `api/` folders (skipping .txt and .md files), leaving it up to you to ensure that `require` will work with your source code.  See the [TypeScript tutorial](http://sailsjs.com/documentation/tutorials/using-type-script) for an example.

PR #3814, opened 2016-08-16
[patch] fixes breaking changes introduced in `include-all` v1.0.1

Ignigena:
Fixes breaking changes introduced in balderdashy/include-all@821ee7848b16872920120f84bf154c331a3939e9

Because the version number in `package.json` uses `^` semver matching, version 1.0.1 is installed by default which breaks functionality on the master branch.


sgress454:
Hi @Ignigena, I think you caught us in the middle of updating the branches, so this should be good now.  Thanks for posting though!

PR #3820, opened 2016-08-25
[patch] Fixes wrong links in contribution guidelines

abisz:
Two links for the documentation guidelines were actually referencing to the test directory of sails.


mikermcneil:
@abisz thanks!

PR #3873, opened 2016-10-30
[proposal] Define a max limit varible

mazyvan:
Change that code in actionUtil file and create a maxLimit var in blueprints.js

### DIFF
--- a/lib/hooks/blueprints/actionUtil.js
+++ b/lib/hooks/blueprints/actionUtil.js
@@ -372,7 +372,9 @@ var actionUtil = {
    */
   parseLimit: function (req) {
     var DEFAULT_LIMIT = req._sails.config.blueprints.defaultLimit || 30;
+    var MAX_LIMIT = req._sails.config.blueprints.maxLimit;
     var limit = req.param('limit') || (typeof req.options.limit !== 'undefined' ? req.options.limit : DEFAULT_LIMIT);
+    if (limit > MAX_LIMIT) { limit = MAX_LIMIT; }
     if (limit) { limit = +limit; }
     return limit;
   },
--- a/lib/hooks/blueprints/actionUtil.js
+++ b/lib/hooks/blueprints/actionUtil.js
@@ -371,10 +371,18 @@ var actionUtil = {
    * @param  {Request} req
    */
   parseLimit: function (req) {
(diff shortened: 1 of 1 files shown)


mikermcneil:
Hey @mazyvan, I think this is a good idea.

Would you have it prefer req.options.maxLimit if available, and camel-case the variable name (since it's not a constant-- I realize why you did it, because the other one is like this, but they really both ought to be camel cased).  e.g.

```js
// An optional ceiling for the `limit`.
// (if exceeded, limit will be set to maxLimit)
var maxLimit;
if (!_.isUndefined(req.options.maxLimit)) {
  maxLimit = req.options.maxLimit;
}
else if (!_.isUndefined(sails.config.blueprints.maxLimit) { 
  maxLimit = sails.config.blueprints.maxLimit;
}
```

Finally, would you also add a test, and include links to the places you're aware of in the docs where this would need to be added?

Thanks!

mazyvan:
Hi @mikermcneil and thank you so much for check this. I did some changes to the code based on what you said. but unfortunately I don't know how to run a test. I mean I know it work haha, but I don't know how to use npm test or stuffs like that.

mikermcneil:
@mazyvan Ok thanks!

@sgress454 so we'd have `maxLimit` in addition to `defaultLimit` (which i think makes sense, and would be straightforward enough to address. And I'd suppose we don't necessarily need to add this sort of a built-in boundary for any of the other criteria clauses....at least not yet.)


In Sails v1, a route pointed at a blueprint action might look something like:

```js
'GET /api/v4/products': {
  action: 'inventory/product/find',
  maxLimit: 100,
  defaultLimit: 25
}
```

or perhaps:

```js
'GET /api/v4/products': {
  controller: 'inventory/ProductController',
  action: 'find',
  maxLimit: 100,
  defaultLimit: 25
}
```

sgress454:
Unfortunately this one timed out -- pretty much all of `actionUtil` has been replaced by `parseBlueprintOptions`.  We do have an example of [how to do `maxLimit` with `parseBlueprintOptions`](http://next.sailsjs.com/documentation/reference/configuration/sails-config-blueprints#?using-parseblueprintoptions), though.

mazyvan:
Nice, thanks

PR #3886, opened 2016-11-10
[patch] Allow policy targeting syntax to use function policies

emckay:
This patch allows routes that looks like:

```
'/foo': [{policy: somePolicyFunction}, {blueprint: 'find', model: 'user'}]
```

(from documentation here: http://sailsjs.org/documentation/concepts/routes/custom-routes#?policy-target-syntax)

Before the patch, policies used in this format had to be strings, as `toLowerCase` was called on them before checking the type. However, the `config/policies.js` file allows policies to be functions as well as strings.

### DIFF
--- a/lib/hooks/policies/index.js
+++ b/lib/hooks/policies/index.js
@@ -231,15 +231,21 @@ module.exports = function(sails) {
         return;
       }
 
-      var policyId = event.target.policy.toLowerCase();
+      var fn;
 
-      // Policy doesn't exist
-      if (!this.middleware[policyId] ) {
-        return Err.fatal.__UnknownPolicy__ (policyId, referencedIn, sails.config.paths.policies);
+      if (_.isFunction(event.target.policy)) {
+        fn = event.target.policy;
+      } else {
+        var policyId = event.target.policy.toLowerCase();
+
+        // Policy doesn't exist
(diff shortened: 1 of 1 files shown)


mikermcneil:
Hey @emckay!  In this case, I think you can do what you're aiming for here using [function target syntax](http://sailsjs.org/documentation/concepts/routes/custom-routes#?function-target-syntax):

```js
'/foo': [
  function (req, res, next) {

    // e.g.
    if (!req.userId) {
      return res.forbidden(new Error('Not authenticated.'));
    }//-•

    return next();

  },
  { blueprint: 'find', model: 'user' }
]
```


But in any case, we definitely could use a better error message for when someone attempts to pass in a function as the RHS of the `policy` key in a route target.  (i.e. that they should just pass in a function for the whole thing instead).  That way, we avoid introducing another way to do this.  If you'd like to change this PR so that it does that error message instead, I'd love to merge it in.

This definitely makes me aware that we need to update the existing docs to cover this use case, so I [did that](https://github.com/balderdashy/sails-docs/commit/b930b53607852e9345528bce943f589e03691541).  If there are other existing docs that could use an update in this area, please let me know or send a PR.

Thanks again!

emckay:
Thank you! The function target syntax worked. I will close this now.

PR #3908, opened 2016-11-30
[patch] fixes global config bug introduced in 44c6d9b

uncletammy:
@sgress454, 

Commit `44c6d9b41e1b1ffd38150788085890c8c2cbe286` doesn't allow for attributes in `config/globals.js` to be commented out.  This patch changes `exposeGlobals.js` to make sure each attribute is defined as well as an acceptable data type before throwing an error.

### DIFF
--- a/lib/app/private/exposeGlobals.js
+++ b/lib/app/private/exposeGlobals.js
@@ -43,7 +43,7 @@ module.exports = function exposeGlobals() {
   // `sails.config.globals._` must be false or an object.
   // (it's probably a function with lots of extra properties, but to future-proof, we'll allow any type of object)
   if (sails.config.globals._ !== false) {
-    if (!_.isObject(sails.config.globals._)) {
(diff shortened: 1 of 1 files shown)


sgress454:
Hi Nick, this is the expected behavior for Sails 1.0 -- if you define `sails.config.globals` at all, you have to define everything.  The default `config/globals.js` file in new v1 apps looks like:
```
module.exports.globals = {
  _: require('lodash'),
  async: require('async'),
  models: true,
  sails: true
};
```

Since we're working towards modularizing everything (`lodash`, `async` and the `orm` hook are all installed at the app level in v1), we were looking at having to make undefined keys default to `false` which would be a breaking change anyway.  We decided in this case it's best to just be as explicit as possible.  It's already been pointed out that those error messages are not as clear as they could be, so I'm about to update them.

Thetzleoo:
GlobalGridline patch was fixed tuesday 1/2017morning personal individual pyramid

PR #3911, opened 2016-12-05
[proposal] Add a command for lifting the sails console in debug mode.

jeff-kilbride:
This proposal adds a `debug-console` command to the Sails CLI, essentially a hybrid of the existing `console` and `debug` commands. This lifts a Sails app in debug mode, allowing interactive debugging from the console.

I use the `sails console` command extensively to test my models, services, etc... It's one of my favorite aspects of Sails development. Having an app lifted with both the console and debugger attached allows me to invoke methods and debug them on the fly. I find it very powerful. I hope others will, too!

It's a fairly simple addition. The only thing I had an issue with was passing the program options (silent, verbose, silly) to the newly spawned debug process. I solved it by using a global variable, but I'm hoping someone else has a better idea...

### DIFF
--- a/bin/sails-debug-console.js
+++ b/bin/sails-debug-console.js
@@ -1,0 +1,48 @@
+#!/usr/bin/env node
+
+
+/**
+ * Module dependencies
+ */
+
+var path = require('path');
+var Womb = require('child_process');
+var CaptainsLog = require('captains-log');
+
+
+/**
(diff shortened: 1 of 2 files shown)


sgress454:
Nice!  Thanks @jeff-kilbride.  Making a minor adjustment to pass the command-line args in without using globals, then merging.

PR #3915, opened 2016-12-07
[patch] Change the test timeout to 10 seconds. It errored with just 4

luislobo:
Change the test timeout to 10 seconds. It errored with just 4. Also, fixed typo


mikermcneil:
@luislobo  👍  thank you!

PR #3916, opened 2016-12-07
[patch] Fixes express 4.x deprecated messages when using send(code, msg)

luislobo:
Since 4.x, send(code, msg) has been deprecated. The current way of sending a code and a message is res.status(code).send(msg). Also, sending just a status code should use sendStatus

Check http://expressjs.com/en/4x/api.html#res.send, http://expressjs.com/en/4x/api.html#res.status and http://expressjs.com/en/4x/api.html#res.sendStatus for more info

### DIFF
--- a/lib/hooks/responses/index.js
+++ b/lib/hooks/responses/index.js
@@ -212,9 +212,9 @@ function _mixin_jsonx(req, res) {
     sails.log.debug('***********************************************************************************');
 
     // Send conventional status message if no data was provided
-    // (see http://expressjs.com/api.html#res.send)
+    // (see http://expressjs.com/en/api.html#res.sendStatus)
     if (_.isUndefined(data)) {
-      return res.status(res.statusCode).send();
+      return res.sendStatus(res.statusCode);
     }
     else if (typeof data !== 'object') {
       // (note that this guard includes arrays)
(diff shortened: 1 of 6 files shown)


luislobo:
ok, fixed!

lib/hooks/responses/index.js
```diff
@@ -212,9 +212,9 @@ function _mixin_jsonx(req, res) {
     sails.log.debug('***********************************************************************************');
 
     // Send conventional status message if no data was provided
-    // (see http://expressjs.com/api.html#res.send)
+    // (see http://expressjs.com/en/api.html#res.sendStatus)
     if (_.isUndefined(data)) {
-      return res.status(res.statusCode).send();
+      return res.sendStatus(res.statusCode);
```
dougwilson: I think this change is incorrect; `res.status(code).send()` is perfectly valid; it sets the status code and sends an empty body. This change causes a body to be sent where it was not before, and the prior code isn't deprecated in any way.

mikermcneil: @dougwilson thanks!

@luislobo I think Scott actually ended up making the change you just made in the other default responses, so it totally makes sense-- but since jsonx is only still included for backwards compat., I think Doug's right and we want to just leave this bit as-is for compatibility's sake

luislobo: Alright, I'll change it back

luislobo:
Is there anything else that needs to be changed? Let me know please

PR #3918, opened 2016-12-07
[patch] Fix reload actions should reload only activated hooks

atiertant:
_.keys(sails.hooks) return all hooks activated or not
_.filter(_.keys(sails.hooks), function(hook) { return sails.hooks[hook]; }) remove false values

### DIFF
--- a/lib/app/reload-actions.js
+++ b/lib/app/reload-actions.js
@@ -32,8 +32,8 @@ module.exports = function reloadActions(options, cb) {
   // Default `hooksToSkip` to an empty array.
   var hooksToSkip = options.hooksToSkip || [];
 
-  // The list of hooks we want to reload is the list of all hooks minus the hooks to skip.
-  var hooksToReload = _.difference(_.keys(sails.hooks), hooksToSkip);
+  // The list of activated hooks we want to reload is the list of all hooks minus the hooks to skip.
+  var hooksToReload = _.difference(_.filter(_.keys(sails.hooks), function(hook) { return sails.hooks[hook]; }), hooksToSkip);
 
   // Clear the actions dictionary.
   sails._actions = {};


lib/app/reload-actions.js
```diff
@@ -32,8 +32,8 @@ module.exports = function reloadActions(options, cb) {
   // Default `hooksToSkip` to an empty array.
   var hooksToSkip = options.hooksToSkip || [];
 
-  // The list of hooks we want to reload is the list of all hooks minus the hooks to skip.
-  var hooksToReload = _.difference(_.keys(sails.hooks), hooksToSkip);
+  // The list of activated hooks we want to reload is the list of all hooks minus the hooks to skip.
+  var hooksToReload = _.difference(_.filter(_.keys(sails.hooks), function(hook) { return sails.hooks[hook]; }), hooksToSkip);
```
mikermcneil: @atiertant Thanks!  This LGTM with a minor tweak:  wouldn't we want to change the comment as follows?

> The list of ~~activated~~ hooks we want to reload is the list of all **activated** hooks, minus the hooks to skip.

@athieriot Also, if you have a moment to add a test for this, I'd appreciate it.


@sgress454 would you mind taking a look at this to confirm, since you were in here last?

Thanks guys!

sgress454:
@mikermcneil @atiertant I'll take a look tomorrow -- as far as I know, hooks that were deactivated by setting `sails.config.hooks.someHook = false` should _not_ be showing up in `sails.hooks` at all, though (it messes with the Sails `inspect` output, for one thing).  There was a bit of time where `pubsub` could be `false` in 1.0, but that's been fixed.  @atiertant I'm assuming you found another scenario with a falsey hook in `sails.hooks`?

atiertant:
@sgress454 for sure ! it was blueprints

sgress454:
@atiertant do you know how that can be reproduced (having `sails.hooks.blueprints === false`)?  When I try lifting with it blueprints disabled in `.sailsrc` or with `sails_hooks__blueprints=false sails lift`, I get `sails.hooks.blueprints === undefined`, which is the expected behavior.

Just to be clear, your PR is appreciated and looks good, but the expectation is that while `sails.config.hooks` can contain falsey values, `sails.hooks` contains only fully-loaded hooks.  That way we don't have to remember to run this sort of filter any time we're looping over `sails.hooks`, and we can assume that if the dictionary contains falsey values, then something went wrong.  So if that expectation is not being met somewhere, I'd like to fix it!

atiertant:
@sgress454 the error was in my code, i thought this was not checked before so you don't need this code

PR #3919, opened 2016-12-07
Use path.join() to avoid any potential issues w/ trailing or leading …

mikermcneil:
…slashes


mikermcneil:
Taking care of this in a separate commit (w/ @sgress454)

PR #3920, opened 2016-12-07
add TODO about removing synchronous file read

mikermcneil:


sgress454:
This looks ok -- verified that in production, it uses cached versions of the locale files:
https://github.com/jeresig/i18n-node-2/blob/b99fd69469d0f014c1a2015800aa206d8ed23998/i18n.js#L347-L349

and the cache is global to the module:
https://github.com/jeresig/i18n-node-2/blob/b99fd69469d0f014c1a2015800aa206d8ed23998/i18n.js#L97

PR #3923, opened 2016-12-08
[proposal] Extend sails command, to allow custom commands

xhallix:
Hi,
this is a proposal to allow custom commands to `sails`. 

So with this change we can do following within our sails app 

```
. approot
+ ...
+ console
   + --- sails.js
+ ...
```

Sails.js
```
var program = require('sails').SailsCommands();

cmd = program.command('custom');
cmd.description('A custom command');
cmd.action(function(){
    console.log('custom programm')
});

if (program.args.length === 0) {
  if(require.main === module) {
    program.usageMinusWildcard();
  }
}
```

This would allow to add some custom logic and group it to the sails commands

```
$ node console/sails.js

  Usage: sails [command]
  Commands:

    version
    lift|l [options]
    new [options] [path_to_new_app]
    generate
    deploy
    console|c
    www
    debug
    help
    ***custom                           A custom command***

  Options:

    -h, --help     output usage information
    -v, --version  output the version number
    --silent
    --verbose
    --silly

```

what do you guys think?

### DIFF
--- a/bin/sails.js
+++ b/bin/sails.js
@@ -162,5 +162,12 @@ program.unknownOption = NOOP;
 program.parse(process.argv);
 var NO_COMMAND_SPECIFIED = program.args.length === 0;
 if (NO_COMMAND_SPECIFIED) {
-  program.help();
+  if(require.main === module) {
+    program.help();
+  }
+}
+
+// make it require able to add custom commands
+module.exports.Commands = function(){
+  return program;
 }
--- a/lib/index.js
+++ b/lib/index.js
@@ -10,6 +10,9 @@ var Sails = require('./app');
 // (maintains legacy support)
 module.exports = new Sails();
 
+// expose it so it can be easily required
+module.exports.SailsCommands = require('../bin/sails').Commands;
+
 // Expose constructor as `.Sails` for convenience/tests:
 // =========================================================
 // To access the Sails app constructor, do:


mikermcneil:
@xhallix Great start!  I'm wondering about a couple of additional things:

1. namespacing (i.e. what about `sails run foo`?)
2. distribution  (how do I share this with my homies?)

xhallix:
@mikermcneil 

1. So you they a custom cmd should have its own namespace?
What is the benefit here when using `sails run foo` instead of `sails foo`?
2. I did not plan to make it shareable, because it will only run in the context of your very own sails app.
I thought about using this as a helper, when you want to start a cron Job, db operations or several custom things in your app which should be done within sails context. Also it help to have comprehensive list of your command available in your Apps.
Php symfonie's command Line approach is doing a great Job and this was inspired by this.
I would not make this to complicate and share it, because most time the cmd operations are related to your pp.

What do you think?

mikermcneil:
> So you say a custom cmd should have its own namespace?

Not necessarily, just wanting to bring it up.  I don't think it would be a good idea to allow core commands to be overridden, because it could get confusing -- but I realize you're not suggesting that here. 

> What is the benefit here when using sails run foo instead of sails foo?

I think the reason it might matter is that it can make it harder to distinguish what commands are part of Sails core and what commands are being introduced by a plugin.  But if custom commands wouldn't be introduced by plugins, then I'm not sure this would matter as much.

That said, there is precedent for the `* run *` usage because of the way NPM scripts work (e.g. `npm run foo`).


> I did not plan to make it shareable, because it will only run in the context of your very own sails app.

Gotcha.  Yeah, I mean I could make something up, but I think it'd be important to see a compelling external use case from for making them shareable -- I'm just bringing it up to talk through it (because these things almost always come up).



---

Here's an example of a `scripts/` folder from a Sails app I've worked on:
image


^ it might be nice to do `sails run compile-terms-of-service` or `sails compile-terms-of-service` -- but I could just as easily run `node scripts/compile-terms-of-service` to achieve the same effect, right? 

---

@xhallix It sounds like the pain points that custom scripts solve are (1) providing a conventional place to put scripts, and (2) providing a baseline that can be improved over time as far as standardizing how CLI opts are declared, etc.  Is that right?  If so, I think we might be able to do this in a formal way that can also optionally take care of loading/lifting sails for you. 

I think I'd prefer to see a solution that can still be kicked off by just running `node scripts/path-to-the-thing.js`, because that makes it easier when using cron etc (i.e. you don't have to learn how to get access to the sails cli in `./node_modules/sails/bin/sails` for production environments where you don't have access to the sails global).   Whether or not you can also kick these scripts off using a top-level custom cmd on the `sails` CLI vs. `sails run name-of-script` is a separate question, I think.

xhallix:
@mikermcneil 

_Namespaces_: I see your point here and you are right. Someone could overwrite a sails command with his own custom command. But since it refers only to a given app, it shouldn't matter much, like you said 👍 

Yes, I'm trying to provide a possibility to have a baseline for all `console` commands.
Image when the app grows, there might be a bunch of `console` commands. So doing `node console/sails` in your app, will give you a comprehensive list about all possible commands within your app. That's the main focus of the proposal.
The benefit is also, that you have access to the `sails` object on `CLI`-tasks, which is nice :)

Maybe `sails new myapp`  could create this logic (`console/sails.js` as posted above) to provide an easy entry point for developers, which would like to add `CLI` functionality.

> I think I'd prefer to see a solution that can still be kicked off by just running node scripts/path-to-the-thing.js, because that makes it easier when using cron etc (i.e. you don't have to learn how to get access to the sails cli in ./node_modules/sails/bin/sails for production environments where you don't have access to the sails global).

You don't need `sails cli in ./node_modules/sails/bin/sails` at any time since `console/sails.js` will represent the same `cli` tasks and everything can run through this on a server

I think it is up to userland to make a script executable via `console/sails` or directly from `node console/antotherCLIScript`. They can always choose how to call it, but I guess most will use `sails newStuff`.

To make things more user friendly I would offer, to provide a new global command, something `sails add newscript` which will automatically adds a `cmd` to `console/sails` and creates a new script file in `console/newscript.js` one can work on.  

If you don't like to see something like that, feel free to close this PR, it's totally fine for me.

Best

luislobo:
I like the idea of having sails commands. But having said that, they need to actually have the whole sails app environment available. For example, in other frameworks, I used to have several batch jobs that were just run using a cron job. All of them used at least the models defined in our app, I can see it useful to have not only models, but also services.
I even would like to be able to "render" views, for example, to send emails using templates, by just rendering it in a view (some ideas).
I guess it could be done in a way where you can actually do a programatically sails.lift...

luislobo:
Sorry for always bringing Yii here, but: http://www.yiiframework.com/doc-2.0/guide-tutorial-console.html

PR #3924, opened 2016-12-09
[patch] Make sails.emit 100% EventEmitter.emit compatible

luislobo:
According to https://nodejs.org/api/events.html#events_emitter_emit_eventname_args, emit should return a boolean: true if the event had listeners, false otherwise.
Also clarified some docs.


luislobo:
ok, fixed!

PR #3950, opened 2017-01-06
[patch] Clean up code in security hook - Updated

luislobo:
Cleans up unused variables, some indentation here and there, and updates README.md. A better explanation should be added about CSRF


sgress454:
Looks good, thanks @luislobo!

PR #3960, opened 2017-01-16
[proposal] Add parseFilter function

cope:
Adding parseFilter function.

The function is identical to the existing parseSort function, only for filter.


sgress454:
@cope I'm not sure I understand what this is for --- especially considering that it's never called.  In any case, `actionUtil.js` is shortly getting the 🔪!

PR #3982, opened 2017-02-12
[patch] Updated `skipper` dependency to use 0.8.0 changes

luislobo:
Updated skipper dependency to use 0.8.0 changes
It has better handling of request type, so that it skips using multipart parsing when not needed


luislobo:
Closing as https://github.com/balderdashy/sails/commit/dd5c9b070e09f6fb9789c7688bec4697548a7c5d fixed it

PR #3983, opened 2017-02-13
[patch] Support for res.status(XXX).ok and res.status(XXX).serverError

luislobo:
Updates responses so that they support using res.status(xxx).ok|serverError . Also show deprecation warnings

For some reason I could not yet figure out, the standard 500 error sets statusCode to 200 so, unit tests fails, I tried to track it couldn't find the reason yet. I'll continue later but I want you guys to check this PR so far.

### DIFF
--- a/lib/router/index.js
+++ b/lib/router/index.js
@@ -408,9 +408,8 @@ Router.prototype.getActionIdentityForTarget = function(target) {
       throw flaverr('E_NOT_ACTION_TARGET', new Error('If target is a dictionary, it must contain an `action` property!'));
     }
 
-    // Bail if the action starts with `somethingController` or contains characters other than letters,
-    // numbers, dashes and forward slashes.
-    if (actionIdentity.match(/^\w+Controller/) || !actionIdentity.match(/^\w+[a-zA-Z0-9\/-]*$/)) {
+    // Bail if the action contains characters other than letters, numbers, dashes and forward slashes.
+    if (!actionIdentity.match(/^\w+[a-zA-Z0-9\/-]*$/)) {
       var didYouMean = '';
       // If the action didn't contain weird characters, make a suggestion by removing "Controller" and
       // replacing dots with slashes.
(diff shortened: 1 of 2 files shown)


mikermcneil:
@luislobo looking really nice!  Esp. appreciate the comment updates and bringing over of the deprecation warnings.  (Not sure what's going on w/ the 200 vs. 500 though either...)

I'm in there w/ blueprints right now, so I'll go ahead and change master to have it do the `return res.status(409).badRequest(err)` thing for the uniqueness errors.  I just merged master into your branch and didn't see any weirdness or conflicts (so far so good!)

sgress454:
Hey @luislobo, all of the updates to comments and deprecation notices look great!  As far as the customizable status codes, now that we have a response generator (`sails generate response`) I'd really rather leave `res.ok()` and `res.serverError()` with their hard-coded 200 and 500 codes.  I understand the intent of making them configurable, but since `res.status()` is sticky, it can lead to situations where you get a status code in the response and it's not clear where it's getting set.  This is one of the problems that the built-in responses are supposed to solve.

@mikermcneil re: the uniqueness errors in blueprints, I think it'd be better to leave it calling `res.badRequest()` and document how to check for uniqueness errors in a custom `badRequest` response (i.e. checking `data.name` and `data.code`).

I'd definitely like to merge this in, if the status codes for the `ok` and `serverError` responses were set back to their hard-coded values!

luislobo:
@sgress454 the issues I see about `ok`, and `serverError` affixing the status to a specific one, is that it is not following Express 4 way of handling responses, like in `.send` and `.json`, where any response you send can override the default status using `status()`. 

I understand that you can override the responses, and that's what we have done from day 1, but, having to code your own response only for sending the status code seems like something very simple to overcome by just setting the default to what you want, 200 and 500 but also respecting the user decision of setting the statusCode. 

I also understand that `ok` and `serverError`, as well as the other responses are Sails API definitions and that does not need to follow Express rules, but, in the end, Sails is overloading Express functionality adding better features and methods.

The other thing is that you cannot say that Sails is a RESTfull API "good player" from the ground up, you have to customize it. Again, that might be something that Sails is not looking forward to, but it would be a plus to just be able to build a REST API with just doing sails new my-api --no-frontend and have everything you need to start building one that respects what everyone is doing and expecting to have when building a REST API.

Again, this are all my points of view and I respect your decisions on this one as you define the API interfaces here. I'm just trying to think about it in the future, where Sails can still be competitive and usable in use cases like microservices (where REST is very used) and not having to explain much on how you can still use it, and why there is no need to go and use featherjs or any other framework to build a simple application/service.

Having said that, until I fix the issue about serverError having a 200 "preset", this PR cannot be merged. 
I'll create one with just the other fixes, and if based on these arguments you still feel like it is not worthy, I can just continue doing what I have been doing, making my own responses.

PR #4001, opened 2017-02-24
[patch] Module loader only js code and config files (and derivatives/dialects)

luislobo:


luislobo:
ok, fixed!

lib/hooks/moduleloader/index.js
```diff
@@ -364,11 +365,11 @@ module.exports = function(sails) {
       }
 
       async.auto({
-        // Load apps from the "api/hooks" folder
```
mikermcneil: Nice catch, thanks!

PR #4038, opened 2017-03-12
eslint

mikermcneil:
Bring in eslint and take care of a few of the most heinous bugs that it surfaced.  This doesn't take care of EVERYTHING though- so the tests will still fail as of the time I'm making this PR.  Once we've taken care of the rest of the warnings (there are rare exceptions where it's appropriate to ignore a rule inline [~~will post an example as a comment momentarily~~ (see below)](https://github.com/balderdashy/sails/pull/4038#issuecomment-285975389)), this PR will be ready to merge.

Here's an initial list of warnings, for convenience:

image

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -5,28 +5,36 @@
   // A set of basic conventions designed to complement the .jshintrc file.
   // For the master copy of this file, see the `.eslintrc` template file in
   // the `sails-generate` package (https://www.npmjs.com/package/sails-generate.)
+  // - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+  // For more information about any of the rules below, check out the relevant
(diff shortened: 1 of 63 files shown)


mikermcneil:
To have eslint ignore a line:

```javascript
var fooBar_baz; // eslint-disable-line camelcase
```

> See https://github.com/node-machine/machine/commit/d0640b58c4d0d9e92d4fbfd811713c2319c28ebb#diff-1522e5213fd7703795b0337442b53911R457 for an example of where the underscores are stylistic and thus this would be convenient/appropriate.  When the underscores are completely needless and aren't stylistic or a property name in some options dictionary we're passing in to an external API, **the best thing to do is just use camel case instead**

luislobo:
:+1:

luislobo:
@mikermcneil Added a PR to this PR https://github.com/balderdashy/sails/pull/4041

PR #4041, opened 2017-03-13
[patch] Corrected the way to actually pass the files to evaluate during eslinting

luislobo:
It specifies the exact folders, and the globbing works on more files, raising the issues to a lot more, unfortunately (or fortunatelly??)

Also, as specific folders are set, node_modules are not evaluated, and it does fail like it fails in Travis, trying to lint node_modules

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,12 @@ env:
 
 before_script:
   - sudo redis-server /etc/redis/redis.conf --port 6380 --requirepass 'secret'
+
+script:
+  # Install latest node, so that eslint can run, as it requires node > 4
   - nvm install stable
   - npm run lint
+  # Revert node to the version being tested
   - nvm use $TRAVIS_NODE_VERSION
+  # Run the test without linting
+  - npm run test-only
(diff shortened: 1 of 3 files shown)


luislobo:
@mikermcneil 
Also note, if you run ...

```bash
node ./node_modules/eslint/bin/eslint accessible bin errors lib --fix
```

... it wil automatically fix what can be fixed according to your .eslintrc config

luislobo:
finally, that eslint version seems to NOT be compatible with 0.10 and 0.12, it uses `const`

particlebanana:
@luislobo we added this line to the `scripts` in some of the adapters to handle the case of earlier versions of node:

```
"pretest": "nodever=`node -e \"console.log('\\`node -v\\`'[1]);\"` && if [ $nodever != \"0\" ]; then npm run lint; fi",
```

This was so we can still have Travis fail if the lint fails so that any PR's will fail for invalid syntax.

https://github.com/balderdashy/sails-postgresql/blob/bdf2b7d7e05b8a85cc0f79918d8add0c88a28298/package.json#L31-L34

luislobo:
@particlebanana yeah, that works, the only drawback is that you don't have linting on older versions but anyway, in my opinion, those versions should not be supported anyway (although code might anyway be compatible with 0.1x)

PR #4043, opened 2017-03-15
[fixes #4042] Allow routes named with \*.Controller*.\i

### LINKED ISSUE #4042: Cannot bind route to controller named with "/.*Controller.*/"
**Sails version**: 0.12.13
**Node version**: 7.7.1
**NPM version**: 4.1.2
**Operating system**: alpine:3.4

<hr/>

I generated a new API using command `sails generate api microController`, it has created model and controller files as expected and RESTful blueprints are working perfectly.

But when I want to add a new custom route:
`'/microcontroller/:id/check': 'MicroControllerController.check'`

I have the following error message in my console:
`error: Ignored attempt to bind route (/microcontroller/:id/check) to unknown controller :: micro.`

MelwinKfr:


sgress454:
By "routes" do you mean "controllers"?  Is this so you can have `api/controllers/MicroControllerController.js`?

MelwinKfr:
Yes I mean "controller" sorry, I already have one named `MicroControllerController` but the thing is that I can't bind a "route" like this: `'/microcontroller/:id/check': 'MicroControllerController.check'` see issue #4042.

It happens because the `controllerId` is normalized two times using `sailsUtil.normalizeControllerId()`. Once before the call to `bindController()` 
[here](https://github.com/balderdashy/sails/blob/0.12/lib/hooks/controllers/to-interpret-route-syntax.js#L92) and a second time inside the function [here](https://github.com/balderdashy/sails/blob/0.12/lib/hooks/controllers/to-interpret-route-syntax.js#L129).

PR #4052, opened 2017-03-21
[patch] Add adapter info to Issues template

luislobo:
What do you think about asking for adapter info and version?
Several times I see that being asked on issues, or that confusion happens because of things working for one adapter but not in an other one.
As always, my 2 cents.

### DIFF
--- a/.github/ISSUE_TEMPLATE
+++ b/.github/ISSUE_TEMPLATE
@@ -21,6 +21,8 @@ NOW THEN: please fill out the following info
 **Sails version**:
 **Node version**:
 **NPM version**:
+**Adapter name**:
+**Adapter version**:
 **Operating system**:
 
 <!--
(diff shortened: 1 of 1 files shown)


sgress454:
@luislobo I like this, but it requires more than just a change to the template, since we'd want this info to be optional.  Right now @sailsbot doesn't do "optional"...

luislobo:
:+1: alright!

sgress454:
@luislobo I like this -- let's modify slightly to:

```
**DB adapter name**: N/A
**DB adapter version**: N/A
```

@sailsbot doesn't do "optional", so we need the "N/A" placeholder for issues where the adapter is not important.

luislobo:
@sgress454 done!

luislobo:
@sgress454 Just a reminder so that you don't forget about this one

PR #4058, opened 2017-03-23
[proposal] Changed the emit('ready') comment, as it now is an important part of Sails

luislobo:
Since Express 4, the router is built in, so middlewares are divided between pre-route and post-route. The way to tell when to do the split is via the ready event

### DIFF
--- a/lib/app/private/initialize.js
+++ b/lib/app/private/initialize.js
@@ -70,8 +70,11 @@ module.exports = function initialize(cb) {
       return cb(err);
     }
 
-    // Fire the `ready` event for backwards-compatibility.
-    // (this will be removed in a future version of Sails!)
+    // Fire the `ready` event
+    // Since Express 4, the router is built in, so middlewares are divided between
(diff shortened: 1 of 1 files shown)


mikermcneil:
@luislobo nice one, thanks!

lib/app/private/initialize.js
```diff
@@ -70,8 +70,11 @@ module.exports = function initialize(cb) {
       return cb(err);
     }
 
-    // Fire the `ready` event for backwards-compatibility.
-    // (this will be removed in a future version of Sails!)
+    // Fire the `ready` event
+    // Since Express 4, the router is built in, so middlewares are divided between
+    // pre-route and post-route. The way to tell when to do the split is via the
+    // ready event
+    // More info in lib/hooks/http/initialize.js:378
```
mikermcneil: for easy reference: 
https://github.com/balderdashy/sails/blob/bdf4e5053c8d0210161b7e62aee25d5e800cbd32/lib/hooks/http/initialize.js#L378-L384

PR #4061, opened 2017-03-30
Feature Request - Optional policy parameters

rene-lindner-isw:
If you are working with lots of different permissions, you end up with a lots of policies that are doing similar things.

Complex permissions could be alot easier, if you could specify Objects in an policy array like `{ 'policyName': parameter }` and make their values available inside the policy via the `req` Object.

This would be optional, existing policies would not break.

For example:

/config/policies.js
```javascript
module.exports.policies = {
  ResourceController: {
    find: [ 'authenticated', { 'myPermissionPolicy': 'ResourceFind' } ],
    create: [ 'authenticated', { 'myPermissionPolicy': 'ResourceCreate' } ],
    findOne: [ 'authenticated', { 'myPermissionPolicy': 'ResourceFindOne' } ],
    update: [ 'authenticated', { 'myPermissionPolicy': 'ResourceUpdate' }  ],
    destroy: [ 'authenticated', { 'myPermissionPolicy': 'ResourceDestroy' }  ]
  },
  ...
};
```

/api/policies/myPermissionPolicy.js
```javascript
module.exports = function( req, res, next ) {
  var requiredPermission = req.policies.myPermissionPolicy.parameter;
  if( !howeverYouCheckYourAuthorizationData( requiredPermission ) ) {
    return res.json( 403, {} );
  }
  next();
};
```

I know there are hooks available, but some completely exchange the policy system, others.have.a('messy').syntax.
And it's not sure if they will be compatible with future sails versions.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -209,6 +209,7 @@ Feature                                          | Proposal                     
  Generate `test/` folder in new Sails apps       | [#2499](https://github.com/balderdashy/sails/pull/2499#issuecomment-171556544)        | Generate a generic setup for mocha tests in all new Sails apps.  Originally suggested by [@jedd-ahyoung](https://github.com/jedd-ahyoung).
  View helper for bootstrapping script tags       | [#3522](https://github.com/balderdashy/sails/pull/3522)                               | Include a view helper for bootstrapping untrusted data from view locals onto the page via script tags in order to expose it to client-side JavaScript. The tricky part is ensuring protection from attempted XSS attacks.
 ~~Improve CORS implementation~~                     | [#3651](https://github.com/balderdashy/sails/pull/3651)                               | ~~Minor changes to the current CORS hooks to better follow the specs/remove inconsistencies.~~
+Optional policy parameters | | Add optional parameters for policies, available via req.
 
 &nbsp;
 &nbsp;


rene-lindner-isw:
ok, fixed!

sgress454:
@rene-lindner-isw Remember that policies can be defined as functions as well, so you can create dynamic policies this way:

```
// config/policies.js
var myPermissionPolicy = require('../api/policies/myPermissionPolicy');

module.exports.policies = {
  ResourceController: {
    find: [ 'authenticated', myPermissionPolicy('ResourceFind') ],
    ...
  }, 
}
```
```
// api/policies/myPermissionPolicy.js
module.exports = function(requiredPermission) {
  return function(req, res, next) {
    if( !howeverYouCheckYourAuthorizationData( requiredPermission ) ) {
      return res.json( 403, {} );
    }
    next();
  }
}
```

I'm not 100% convinced that formalizing this is worth the extra code in Sails...I will leave it to the will of the community to decide!

Also -- your specific example is probably simplified for demonstration purposes, but in case it's your real-world case I have to point out that you don't need dynamic policies at all; you can use `req.options.action` inside the policy to determine which action is currently running!

rene-lindner-isw:
Hm I don't like the idea of using `req.options.action` inside the policy to determinate which permission to use. Imho the information how each action is protected should be comprehensible by just looking at the `config/policies.js`.

However the other example with the function as policy looks really promising, thanks alot for the hint!
It's absolute enough for my usecase, and imho it's also very clean and comprehensible.

From my side, this can be closed.

rene-lindner-isw:
Maybe an example like this should be added to the docs?
http://sailsjs.com/documentation/concepts/policies
After "Built-in policies", "Dynamic policies" ?

PR #4067, opened 2017-04-10
Update loadHooks.js

tacshooter:
Actual error message output for "A hook('hookname') failed to load..." This is one of the hardest things to debug in a sails app and there's no reason at all for this error not to be included here.

### DIFF
--- a/lib/app/private/loadHooks.js
+++ b/lib/app/private/loadHooks.js
@@ -171,6 +171,7 @@ module.exports = function(sails) {
           _.each(hookTimeouts, function(hookTimeout) {clearTimeout(hookTimeout);});
           if (id !== 'userhooks') {
             sails.log.error('A hook (`' + id + '`) failed to load!');
+            sails.log.error(err);
           }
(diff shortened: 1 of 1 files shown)


sgress454:
@tacshooter Can you show an example where the error doesn't already get logged to the console?  [This line](https://github.com/balderdashy/sails/pull/4067/files#diff-bae0b3503095a5968ed03a9fb966ff0aL176) should cause it to bubble up, which in turn should cause Sails to fail and throw that error.  If we implement your change, we'll end up with a lot of double error logs.  But if there's a case where the error is getting swallowed, I'd rather find and patch that.

tacshooter:
@sgress454 - I usually see it when there's a problem with the datastore connection or with one of the models. The console at startup just says "A hook orm failed to load" and shows no other detail. So I've just made this change in Sails so that I can see the actual error. This comment is a few months old, so it'd be difficult to get back to the state where I was seeing this happen a lot. I can tell you that all the times I've ever seen this happen before making my own change, I never once saw the error get logged to the console.

Issue #1150, opened 2013-11-28
Enable Express compression in Sails.js by default


pierredewilde:
In `config/express.js` file:

```
module.exports.express = {
    customMiddleware: function (app) {
        var express = require('../node_modules/sails/node_modules/express');
        app.use(express.compress());
    }
};
```

It will speed up Sails.js by compressing all Express responses.

mattcodez:
@pierredewilde Is there a simple way to enable compression only for certain methods? Given the discussion in this thread, off by default on JSON sounds appropriate, so I'd like to enable for a particular method I have. I tried to setup using policies, connection hangs and I'm guessing it's because policy middleware runs before the application method, and I would need this to run after right?

pierredewilde:
Dunno.

animedbz16:
Or you could use the buildProd grunt task and host static assets behind nginx so that your node server doesnt even have to handle serving static assets since if it had to handle serving static assets this means it can't be doing other things that it actually has to do such as handling API requests and interfacing with the database.

Leave static assets to be handled by nginx for production. The buildProd grunt task already handles concatination, minification and uglification of statics assets into a `www` directory which a simple nginx config can point to handle serving those assets and routing all other requests to the backend node server.

PR #4081, opened 2017-04-22
[proposal] - Option for compressing assets

noway:
As specified in https://github.com/balderdashy/sails/issues/1150, compression on the fly is too costly. 

Compression could be implemented only for the assets though and compressed files could be cached. This would involve an option which would compress files in .tmp folder and then sails http server would specifically serve only them when the client is requesting static files.


animedbz16:
Or you could use a real web server like nginx to handle serving static
files so your node server doesn't have to at all.

noway:
Yes, but that way it complicates cloud deployment a lot. Instead of deploying single app you'd be deploying app+nginx server with it, or using cdn which might be not appropriate in your case.

animedbz16:
You are concerned with sails / nodes performance in context of serving up static assets and thus request caching of compressed assets to serve them out faster.

Node runs as a single process, so when you have node resolving requests such as sending out static assets, it means that it can not be processing other requests that it must actually do like handling API requests and interfacing with a database.

If you are concerned with performance of handling static assets, then you should stand up nginx as it is designed to do this much more efficiently since it is coded in C and it is the recommended way to deploy to production.

noway:
Surely node can stream multiple streams at once, events are fired on each io wait I believe.

> and taking your build www to host behind nginx is not complicated

In cd environment, it is. Rather then deploying single build you'd be deploying bits and pieces to nginx instances/cdn/etc. Much better I think is to leave handling of static files to sails and then cache them on load balancer level.

PR #4129, opened 2017-06-22
[patch] Enhancements to support Redis on Heroku

davidjconnolly:
Given the verbose log level of the Redis error on L312, I'm assuming the errors on L320 and L327 were meant to be logged with a verbose log level as well. We're on Heroku and are getting hundreds of these errors which is flooding our Sentry logging as the Redis connections timeout every 300 seconds.

Another potential implementation would be to move the errors to an else block after the check for onRedisDisconnect and onRedisReconnect functions if the intent is for the onRedis... to provide custom overriding rather than supplemental logic.

Also fixing what I suspect is a typo on L324 where currently the logic checks for an onRedisDisconnect definition but then fires onRedisReconnect.

Lastly, changing _.extend ->_.merge on L440 to allow (in our case)  `saveUninitialized` to be set to false.  We deployed to production and very quickly filled our Heroku Redis store with uninitialized sessions.

### DIFF
--- a/lib/hooks/session/index.js
+++ b/lib/hooks/session/index.js
@@ -317,14 +317,14 @@ module.exports = function(app) {
                   if (_.isFunction(app.config.session.onRedisDisconnect)) {
                     app.config.session.onRedisDisconnect();
                   }
-                  app.log.error('Redis session server went offline...');
+                  app.log.verbose('Redis session server went offline...');
                   // If a disconnected client comes back, say something and run any custom logic
                   // that was provided for this occasion.
                   err.connection.once('ready', function() {
                     if (_.isFunction(app.config.session.onRedisDisconnect)) {
                       app.config.session.onRedisReconnect();
                     }
-                    app.log.error('Redis session server came back online...');
+                    app.log.verbose('Redis session server came back online...');
                   });
                 }
               }
--- a/lib/hooks/session/index.js
+++ b/lib/hooks/session/index.js
@@ -318,10 +318,10 @@ module.exports = function(app) {
                     app.config.session.onRedisDisconnect();
                   }
                   app.log.verbose('Redis session server went offline...');
-                  // If a disconnected client comes back, say something and run any custom logic
+                  // If a disableconnected client comes back, say something and run any custom logic
                   // that was provided for this occasion.
                   err.connection.once('ready', function() {
-                    if (_.isFunction(app.config.session.onRedisDisconnect)) {
+                    if (_.isFunction(app.config.session.onRedisReconnect)) {
                       app.config.session.onRedisReconnect();
                     }
                     app.log.verbose('Redis session server came back online...');
--- a/lib/hooks/session/index.js
+++ b/lib/hooks/session/index.js
@@ -318,7 +318,7 @@ module.exports = function(app) {
                     app.config.session.onRedisDisconnect();
                   }
                   app.log.verbose('Redis session server went offline...');
-                  // If a disableconnected client comes back, say something and run any custom logic
+                  // If a disconnected client comes back, say something and run any custom logic
                   // that was provided for this occasion.
                   err.connection.once('ready', function() {
                     if (_.isFunction(app.config.session.onRedisReconnect)) {


sgress454:
@davidjconnolly Re: [`_.extend` vs `_.merge`](https://github.com/balderdashy/sails/pull/4129/commits/ad87f6837d266e80871a6edd4432ff02636205a4#diff-30772e207fff1c62d9567fb36d10e453L440) -- I don't see why this change is necessary.  Setting `resave` and `saveUninitialized` to `false` in my `config/session.js` or `config/production.js` works fine for me with the current code.  How are you configuring your session store?

davidjconnolly:
@sgress454 we're configuring it in our config/session.js as follows (using our fork https://github.com/opencare-admin/sails):
```
'use strict';

var session = {
  // Redis options:
  adapter: 'redis',
  ttl: 2592000, // 30 days session expiry
  db: 0,
  prefix: 'sess:',

  // Express options:
  name: 'dc.sid',
  secret: '<removed>',
  resave: false,
  saveUninitialized: false,
  unset: 'destroy',
  cookie: {
    secure: true,
    maxAge: 30 * 24 * 60 * 60 * 1000
  }
};

module.exports.session = session;
```

I'm trying to remember now exactly what the issue was, I believe that the source object was overwriting the destination `saveUninitialized` property.

sgress454:
@davidjconnolly Hm, I installed your fork and changed `_.merge` back to `_.extend`, and added the session config from above, and got the same result when I log `opts` -- `resave` and `saveUninitialized` both `false` as expected.  Perhaps there was another issue that got fixed simultaneously (like a conflicting config in `config/env/production.js`)?  

The typo you caught definitely needs to be merged.  The log verbosity we'll consider, but can you say a bit more about why your session is disconnecting/reconnecting so much?  That is, is there an argument to be made that in general, these errors are not serious enough to always show up?

davidjconnolly:
@sgress454 That's weird, I just tried it locally as well and had no issue, @vadim-zverugo do you remember why we needed to change _.extend to _.merge?  I seem to remember needing this change in order for saveUninitialized to stay false but I tried it both ways with a console log of `opts` on L445 and it's preserved as false in both cases.

With respect to the log verbosity, it seems we've resolved our issue with the main API redis connections timing out as they're kept alive by all the traffic.  I suspect we had some issues with orphaned dyno's.  However, our job scheduler spins up sails which leaves an un-used redis connection that is timing out harmlessly after the default heroku timeout of 5 minutes.

I definitely see the argument for wanting to default to log.error for the general case but it was frustrating that we couldn't override it.  What if I instead put an else on L319 and L326 so that the default behaviour is to throw an error but if the `onRedisDisconnect/Reconnect` override functions are set, the alerting and any other logic is delegated to them?  I'll back out the _.extend change too if we can't find a good reason for keeping it.

Thoughts?

sgress454:
>  What if I instead put an else on L319 and L326...

I like this!  I'll make a similar change in [sails-hook-sockets](https://github.com/balderdashy/sails-hook-sockets/blob/master/lib/ensure-redis-connection.js#L89-L95).  

> I'll back out the _.extend change too if we can't find a good reason for keeping it.

Sounds good.  It'll be good to have y'all back in the fold.  Every time someone uses a forked Sails in production, a puppy dies.

vadim-zverugo:
@sgress454, @davidjconnolly: Agree, `_.extend` works well and overwrites `resave` and `saveUninitialized` properties. Looks like I had another issue with a custom config.

Once Redis/log patches are confirmed and applied, I'll be happy to upgrade to Sails v.1.0.0-37.

And one more question about `sails-hook-orm`. Currently, in order to work with Sails v.1.0, we should specify in our `package.json` something like `"sails-hook-orm": "git+https://github.com/balderdashy/sails-hook-orm.git"`. When are you planning to publish this package in NPM?

sgress454:
@vadim-zverugo re: sails-hook-orm, you should be able to set your dependency to:

```
sails-hook-orm": "^2.0.0-20"
```

That's the semver range for the Sails 1.0-compatible hook.

davidjconnolly:
Thanks @sgress454 & @vadim-zverugo, feedback addressed!  Let's resurrect some puppies ;)

sgress454:
Thanks for the great work @davidjconnolly and @vadim-zverugo.  We expect to be publishing a new 1.0 version next week.

sgress454:
@davidjconnolly @vadim-zverugo Sails v1.0.0-37 was published yesterday, sorry for the delay!

PR #4219, opened 2017-10-19
[patch] alwaysAllow function name should not be neverAllow

luislobo:
Changed alwaysAllow function name to be alwaysAllow


sgress454:
Ayup.  Thanks @luislobo!

PR #4303, opened 2018-02-02
[patch] Use node module resolution properly

Swatinem:
For quite a few cases, sails does not use the default node module resolution algorithm, but builds explicit paths in `${appPath}/node_modules/`, which is not nice.

### DIFF
--- a/lib/hooks/session/index.js
+++ b/lib/hooks/session/index.js
@@ -360,7 +360,7 @@ module.exports = function(app) {
             if (_.isString(app.config.session.adapter)) {
 
               try {
-                SessionAdapter = require(path.resolve(app.config.appPath, 'node_modules', app.config.session.adapter));
+                SessionAdapter = require(app.config.session.adapter);
               }
               catch(rawRequireErr) {
 


mikermcneil:
@Swatinem Why do you feel that this is not nice?

Swatinem:
We have multiple sails apps inside a mono-repo, and would like to hoist some dependencies to the toplevel.
We worked around this problem anyway by just creating symlinks where needed.

tomasfrancisco:
I might have the same problem. 

I'm trying to set up a mono repo with yarn workspaces and I'm getting the following error when I try to run sails:
```
Failed to lift app: TypeError: Cannot read property 'readJson' of undefined
    at Object.module.exports.bootstrap (/Users/tomasfrancisco/Documents/dev/tap-water/api/config/bootstrap.js:53:8)
```

@Swatinem Can I ask you what exactly did you do to fix it?

Swatinem:
Well you can:
* use the `nohoist` feature of recent `yarn > 1.5` to exclude certain things from hoisting
* manually create symlinks from `/packages/$FOO/node_modules/$BAR -> /node_modules/$BAR`

We went the symlink route because yarn still hoists your workspace packages themselves, and we needed those because some of them were sails hooks.

So you want to `nohoist`:
* connect-redis
* sails-mysql (or equivalent)
* anything that is a sails hook. Sails basically enumerates every package in `./node_modules/…` and looks at the `package.json` to figure out if its a hook or not (please correct me if I‘m wrong)

PR #4375, opened 2018-04-26
[proposal] Add support for Vuejs templates

Melchyore:
Hello!
Since consolidate doesn't support Vuejs templates, I think it should be useful for a lot of web developers to be able to use this templates as it makes it much easier to work on the client-side while using Sails global variables inside. The express-vue module can solve that, but it uses a .renderVue() function which is not native .view() Sail's function and so we can't use Sail's globale variables inside Vuejs templates.


mikermcneil:
@Melchyore I can dig it -- not sure how I feel about `.vue` files, but at the very least I'd like there to be a clear path to setting this up documented, especially since we're using vue.js in the "Web App" template now.

I think first step here is a [tutorial](https://github.com/balderdashy/sails-docs/new/master/tutorials) on the subject -- specifically the simplest way to quickly set this up in a Sails app.  Once we've got that, it'll be a lot easier to see how to go about planning a plugin-powered or core implementation

Melchyore:
I'm happy to read that because with this two big tools, we can achieve something great in a trice!

Well there's a lot of possibilities but I think the best way to achieve this is to take advantage of the getRenderFn function where we'll be able to configure the render process (e.g ssr or csr).

PR #4379, opened 2018-04-27
[proposal] Add route() function (like Laravel)

Melchyore:
Hello!

Let's imagine a scenario where we have a list of users and clicking on user's name will redirect to its profile. So we could have something like that :
In config/routes.js :
```js
// Some routes
'GET /user/:id': {
  controller: 'UserController',
  action: 'show'
// Some routes
}
```
Then, in the template file (e.g. users.ejs) :
```ejs
<% _.forEach(users, (user) => { %>
  <a href="<%= sails.getUrlFor('UserController.show').replace(':id', user.id) %>"><%= user.username %></a>
<% }) %>
```

As you can see, it's too long to write the route and it's not the best way to replace the parameter with its value.

As a solution :
- Add `name` property inside 'GET /user/:id' object with 'get_user' as value.
- In sails.router, add a 'namedRoutes' property which will be an Object whose keys will be the names of the routes and the values ONLY the URLs of the routes, something like that :
```js
get_user: '/user/:id'
```
- Add a global function route(routeName, ...values) which will replace all the named parameters with their values and returns the URL string. So, instead of writing `<a href="<%= sails.getUrlFor('UserController.show').replace(':id', user.id) %>">`, we will write `<a href="<%= route('get_user', user.id) %>">` in the users.ejs template.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -30,7 +30,7 @@ Feature                                          | Proposal                     
  :---------------------------------------------- | :------------------------------------------------------------------------------------ | :----------------------------------------------------------------------------------------------------------
  Allow select/omit clauses when populating a singular association | https://trello.com/c/yM9WPxzr/107-waterline-fs2q-tolerate-a-subcriteria-being-provided-to-populate-for-a-singular-associations-but-only-if-it-exclusively-contains | Don't throw an error if these clauses are included in a `populate` for a singular association (but still error if actual "where" criteria are used)
  Generate `test/` folder in new Sails apps       | [#2499](https://github.com/balderdashy/sails/pull/2499#issuecomment-171556544)        | Generate a generic setup for mocha tests in all new Sails apps.  Originally suggested by [@jedd-ahyoung](https://github.com/jedd-ahyoung).
-
+Add route() function (like Laravel) | | Generate a URL with named parameters replaced by their values automatically.
 
 
 <!--


Melchyore:
I wrote a hook to do this.
```js
let detectVerb = require('../../../node_modules/sails/lib/util/detect-verb')

module.exports = function defineRouteHook(sails) {
  return {
    configure () {
      let routes = sails.config.routes

      sails.router.namedRoutes = {}

      _.forIn(routes, (options, route) => {
        if (_.isObject(options) && _.has(options, 'name')) {
          let url = detectVerb(route.toLowerCase()).path

          if (url.startsWith('/')) {
            url = url.substr(1)
          }

          sails.router.namedRoutes[options.name] = url
        }
      })
    },

    /**
     * Runs when a Sails app loads/lifts.
     *
     * @param {Function} done
     */
    initialize (done) {
      sails.route = (routeName, ...parametersValues) => {
        let route = sails.router.namedRoutes[routeName]

        if (!route) {
          throw new Error('No named route : ' + routeName + '\n Maybe you missed the name field?')
        }

        let parameters = route.match(new RegExp(/:[a-zA-Z_]+/, 'g'))

        if (parameters) {
          if (parameters.length !== parametersValues.length) {
            console.warn('Length of named parameters (' + parameters.length + ') is not the same as the length of values ' + parametersValues.length + ' ')
          }

          for (let key of parameters.keys()) {
            route = route.replace(parameters[key], parametersValues[key])
          }
        }

        return route
      }

      sails.on('router:before', () => {
        sails.router.bind('all /*', (req, res, next) => {
          res.locals.route = sails.route

          return next()
        })
      })

      return done()
    }
  }
}
```

mikermcneil:
> I wrote a hook to do this.
👍 thanks for posting!

Melchyore:
@mikermcneil No problem
I'm looking to add more options like grouping routes, parameters validation, etc.

PR #4394, opened 2018-05-06
Emit an event on model's error

Melchyore:
When inserting/updating/finding/deleting, if there's an error (not ORM error), emit un event "model:error" with the Error Object as argument for the event callback.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -30,6 +30,7 @@ Feature                                          | Proposal                     
  :---------------------------------------------- | :------------------------------------------------------------------------------------ | :----------------------------------------------------------------------------------------------------------
  Allow select/omit clauses when populating a singular association | https://trello.com/c/yM9WPxzr/107-waterline-fs2q-tolerate-a-subcriteria-being-provided-to-populate-for-a-singular-associations-but-only-if-it-exclusively-contains | Don't throw an error if these clauses are included in a `populate` for a singular association (but still error if actual "where" criteria are used)
  Generate `test/` folder in new Sails apps       | [#2499](https://github.com/balderdashy/sails/pull/2499#issuecomment-171556544)        | Generate a generic setup for mocha tests in all new Sails apps.  Originally suggested by [@jedd-ahyoung](https://github.com/jedd-ahyoung).
+ Emit an event on model's error | | Emit an event on create/update/find/delete queries if there'is an error.
 
 
 


Melchyore:
ok, fixed!

mikermcneil:
@Melchyore what's the use case you're envisioning for this?

I see how this could fit in with error reporting/alerts, and I'm open to something like it eventually.  Today, I usually see the alerts/reporting use case handled with built-in alerts features in tools like Papertrail, but I know those tools aren't always an option for every project.  If there's another use case here that I'm missing, let me know!

If we do add a future like this, rather than another type of event, I think we might be better served by an `afterQueryError` lifecycle callback, configured as a top-level model setting:

```js
afterQueryError: async (err, s2q)=>{
  // err => an Error instance
  // s2q => a fully-normalized "stage 2 query", as defined by waterline

  // If an uncaught exception is thrown or unhandled Promise rejection occurs from within
  // this lifecycle callback, that additional error is logged using the built-in sails logger
  // (i.e. `sails.log.error()`).  Keep in mind the original error (`err`) will already have been
  // handled as usual at this point, so there'd be no reason to re-handle it.  In other words,
  // the point of this lifecycle callback would be for _additional_ logic that occurs on-error.
}
```

(There's also an argument to be made for an interruptible `beforeQueryError` LC, but I'm a bit more skeptical because of the potential issues that could cause.)

Melchyore:
@mikermcneil as you know, Sails doesn't support custom validation messages. So, to validate requests, we must check the error code (e.g. E_UNIQUE) then return the appropriate message, but that's is redundant if there's a lot of requests. So why not automate the process?
If ther's an event on validation error, a hook could intercepting it then checking if the message exists in the model and binds it to the res.locals Object.

`/api/models/Foo.js` :
```js
module.exports = {
  attributes: {
    bar: {
      type: 'string',
      unique: true
    }
  },
  validationErrors: {
    bar: {
      E_UNIQUE: 'The attribute bar is unique blabla.'
    }
  }
}
```

`validate.js` in the initialize function :
```js
sails.on('model:error', err => {
  if (err) {
    res.locals.errors.push(sails.models[err.modelIdentity].validationErrors[err.code])
  }
})
```
The main purpose of this PR is to avoid writing something like :
```js
if (err) {
  if (err.code === 'E_UNIQUE') {
    errors.push('The attribute bar is unique blabla.')
  } else if (err.code === 'SOME_ERROR_CODE') {
    errors.push('Message for SOME_ERROR_CODE')
  }
}
```

If your suggesion can do the same thing, that would be awesome ! :o

mikermcneil:
Ah, I see.  So what I suggested wouldn't let you do this.  In order to intercept all errors across the board, you'd need a `before`-style lifecycle callback.  I'm open to it-- not sure how I feel about it being asynchronous, but worth a think


side note, re:

> …The main purpose of this PR is to avoid writing something like…


Just a side note-- you might check out [.intercept()](https://sailsjs.com/documentation/reference/waterline-orm/queries/intercept) -- it can save a bunch of time and duplication (e.g. with special exit signals: https://github.com/mikermcneil/ration/blob/276df86210f78dc065764b94f9a2ca5840b91767/api/controllers/entrance/signup.js#L81   -- combine that with a custom response and sounds like that might take care of it)

Melchyore:
Oh thank you, I will give it a try!
May I ask you what are you meaning by "custom response"?

mikermcneil:
@Melchyore https://sailsjs.com/documentation/concepts/extending-sails/custom-responses

Melchyore:
@mikermcneil thanks!

PR #4422, opened 2018-05-24
[proposal] Cache node_modules

DanielRuf:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,3 +30,7 @@ env:
   TEST_REDIS_SESSION=true
 
 before_script: sudo redis-server /etc/redis/redis.conf --port 6380 --requirepass 'secret'
+
+cache:
+  directories:
+    - node_modules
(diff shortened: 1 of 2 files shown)


mikermcneil:
@DanielRuf Only problem with this is that it doesn't catch breakage that could come from upstream dependencies (e.g. if a buggy patch release of `accepts` was published: https://github.com/expressjs/express/blob/3ed5090ca91f6a387e66370d57ead94d886275e1/package.json#L34)

DanielRuf:
> @DanielRuf Only problem with this is that it doesn't catch breakage that could come from upstream dependencies (e.g. if a buggy patch release of accepts was published: expressjs/express:package.json@3ed5090#L34)

I'm not sure if I understand correctly. By default Travis created checksums (eg of lockfiles) and caches the directory and only updates the cache if files have changed.

The line points to some other package.

PR #4434, opened 2018-05-31
[proposal] Add new feature proposal

ThatNerdyPikachu:
Basically a more stripped down version of WebApp, so people who don't want to use email or billing don't have to tediously remove those features with various degree of knowledge.

Maybe add basic OAuth2 support to it maybe?


mikermcneil:
@ThatNerdyPikachu Makes sense- thanks for bringing this up.  If someone has time to start on this, I'd be happy to review a pull request (I think we'll want to get it completely done in one PR though)

We could overload the existing `--without` option

PR #4505, opened 2018-09-06
[patch] Replace uid-safe to nanoid

ai:
`uid-safe` is a random string ID generator. Sails uses it in session ID generator.

This PR replaces `uid-safe` to [`nanoid`](https://github.com/ai/nanoid)

> Why do we need it?

1. Nano ID is [10% faster](https://github.com/ai/nanoid#benchmark)
2. Nano ID has no dependencies compare to 1 dependency of `uid-safe`
3. Nano ID has automatic tests that [distribution is flat](https://github.com/ai/nanoid/blob/master/test/index.test.js#L28-L52) and everything is secure
4. We have the cool logo 😄 and [good documentation](https://github.com/ai/nanoid/blob/master/README.md). Even a special website to calculate [ID length](https://zelark.github.io/nano-id-cc/).

> Is it a big project or some unknown project that we should not care about

Nano ID is not just new unknown project. It has [1M+ downloads per month](https://npm-stat.com/charts.html?package=nanoid).

Docs was translated to Chinese. It has ported to 13 langauges](https://github.com/ai/nanoid#other-programming-languages).

> Can we trust this project?

Nano ID was created by the author of PostCSS and Autoprefixer. The community can trust me 😄.

Nano ID right now is used by many users. For instance, it is used by [Zeit’s Next.js](https://github.com/zeit/next.js/commit/918e0a6e3230ef0e33a5698cd09aa274cb0a1fec).

> Is it safe?

Yeap, Nano ID uses hardware randome generator (same as `uid-safe`). We really care about security, we have automatic tests for security. For instance, we tests flat distribution (it guarantees unpredictability) [in Node.js](https://github.com/ai/nanoid/blob/master/test/index.test.js#L28-L52) and [browsers](https://github.com/ai/nanoid/blob/master/test/demo/index.js).

@mikermcneil to be honest, it is not a big PR which dramatcally change everything. It will just a very little increase perfomance and remove one dependency. But in other hand, it is small cute change without any risks. During PostCSS maintaince, I very like this type of PR 😎

### DIFF
--- a/lib/hooks/session/index.js
+++ b/lib/hooks/session/index.js
@@ -9,7 +9,7 @@ var flaverr = require('flaverr');
 var Redis = require('machinepack-redis');
 
 // (this dependency is just for creating new cookies)
-var uid = require('uid-safe');
+var nanoid = require('nanoid');
 
 // (these two dependencies are only here for sails.session.parseSessionIdFromCookie(),
 //  which is only here to enable socket lifecycle callbacks)
@@ -480,7 +480,7 @@ module.exports = function(app) {
 
     generateNewSidCookie: function (){
 
-      var sid = uid.sync(24);
+      var sid = nanoid(32);
       var signedSid = 's:' + signCookie(sid, app.config.session.secret);
       var cookie = stringifyCookie(app.config.session.name, signedSid, {});
       return cookie;
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
     "merge-defaults": "0.2.1",
     "merge-dictionaries": "1.0.0",
     "minimist": "0.0.10",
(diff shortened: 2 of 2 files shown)


ai:
@mikermcneil Travis CI has some problem with Node 9.0. Maybe we can just start it again? (I have no access)

ai:
@rachaelshaw @mikermcneil in Nano ID 1.2.4 we found [a way](https://github.com/ai/nanoid/pull/82) to make Nano ID, even more, faster than `uid`:

```
nanoid                400,771 ops/sec
uid.sync              357,449 ops/sec

nanoid/async           85,251 ops/sec
uid                    76,991 ops/sec
```

PR #4561, opened 2018-12-14
[patch]Update package.json changed the i18n-2 version

MatteoAzzarelli:
Update version of i18n-2 from 0.6.3 to 0.7.3 to solve the following problem derivates from dependencies of this package:
deprecated sprintf@0.1.5: The sprintf package is deprecated in favor of sprintf-js.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "express-session": "1.15.6",
     "flaverr": "^1.9.0",
     "glob": "7.1.2",
-    "i18n-2": "0.6.3",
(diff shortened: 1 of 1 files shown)


rachaelshaw:
@MatteoAzzarelli thanks for getting rid of that pesky deprecation warning! I tested a new app with the upgraded dependency and it worked fine, so this looks good to merge

MatteoAzzarelli:
It's a big satisfactions! Thanks

PR #6543, opened 2019-05-31
[fixes #6541]. omit prefix to avoid doubling up when key gets set

### LINKED ISSUE #6541: Redis sessions with custom prefix include the prefix twice in the key
**Node version**: 8.15
**Sails version** _(sails)_: 1.2.2
**ORM hook version** _(sails-hook-orm): 2.1.1
**Sockets hook version** _(sails-hook-sockets)_: 1.5.5
**Grunt hook version** _(sails-hook-grunt)_: 1.0.8
**DB adapter & version** _(e.g. sails-mysql@5.55.5)_: sails-postgresql@1.0.2
**connect-redis version**: sailshq/connect-redis@3.2.1
<hr/>

Within our config:session.js file, we have defined a custom prefix for our redis key as such. 

```
module.exports.session = {
  prefix: ‘foo:’
  ...
};
```

When we initialize a new session, the key being stored in redis is actually “foo:foo:key_value_here.”

While this is fine for a single app, we share a session between multiple applications, so the session is immediately replaced by another sails app who’s looking for “foo:key_value_here” and is not finding it.

travispwingo:
Fix for issue: https://github.com/balderdashy/sails/issues/6541.

The prefix was being included twice in the redis store.

### DIFF
--- a/lib/hooks/session/index.js
+++ b/lib/hooks/session/index.js
@@ -302,7 +302,7 @@ module.exports = function(app) {
             // Create a Redis connection manager.
             Redis.createManager({
               connectionString: url,
-              meta: _.omit(app.config.session, ['host', 'port', 'pass', 'db', 'url', 'adapter']),
+              meta: _.omit(app.config.session, ['host', 'port', 'pass', 'db', 'url', 'adapter', 'prefix']),
               // Handle failures on the connection.
               onUnexpectedFailure: function(err) {
                 // If Sails is already on the way out, ignore the Redis issue.


travispwingo:
@johnabrams7 has this been discussed/reviewed yet for merge?

johnabrams7:
@travispwingo Sorry for the delay, we reviewed and confirmed this resolves the duplicate prefix for the redis session key in redis-cli console. Tested it with multiple simultaneous sessions without issues. Will have this included with a sails update shortly.

PR #6847, opened 2019-08-24
[fixes #5630] Update destroy blueprint action so that `afterDestroy` lifecycle callbacks will still run by default

### LINKED ISSUE #5630: Blueprints action destroy is not called with meta
**Node version**: 8.16.0
**Sails version** _(sails)_: 1.2.2
**ORM hook version** _(sails-hook-orm)_: 2.1.1
**DB adapter & version** _(e.g. sails-mysql)_: 1.0.1
<hr/>

From documentation:
> The afterDestroy lifecycle callback will only be run on .destroy() queries that have the fetch meta flag set to true.

Blueprints action destroy don't include meta object so afterDestroy lifecycle callback don't run.

rachaelshaw:
Updated the destroy blueprint action so that the meta query options from `parseBlueprintOptions` are passed into the actual call to `.destroy()`, since [`afterDestroy` only runs if `.meta({fetch: true})` is included](https://sailsjs.com/documentation/concepts/models-and-orm/lifecycle-callbacks#?lifecycle-callbacks-on-destroy). This makes the `afterDestroy` lifecycle callback still run by default when using the destroy blueprint action.

Fixes #5630

### DIFF
--- a/lib/hooks/blueprints/actions/destroy.js
+++ b/lib/hooks/blueprints/actions/destroy.js
@@ -53,10 +53,11 @@ module.exports = function destroyOneRecord (req, res) {
 
     // (Note: this could be achieved in a single query, but a separate `findOne`
     // is used first to provide a better experience for front-end developers
-    // integrating with the blueprint API out of the box.  If we didn't need
-    // or care about that, we could just use `.meta({fetch: true})` when calling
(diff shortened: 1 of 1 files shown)


rachaelshaw:
@sailsbot my apologies, your grace

Issue #6838, opened 2019-08-13
Under certain conditions, disabling CSRF for a route doesnt work (Debugging included)


kinsi55:
**Node version**: 8
**Sails version** _(sails)_: 1.2.3
<hr/>

In https://github.com/balderdashy/sails/blob/b3afed7a6c67ba61b5202cab5954a388f05617c2/lib/hooks/security/csrf/index.js#L32

Due to the Path being lowercased, if you have a RegEx pattern with uppercased latters, it will then fail to properly match against a given request URL further down (https://github.com/balderdashy/sails/blob/b3afed7a6c67ba61b5202cab5954a388f05617c2/lib/hooks/security/csrf/index.js#L99), and thus always return Forbidden.

I'm not sure if there would be any regressions from removing the .toLowerCase() but certainly this shouldnt be desired behaviour.

sailsbot:
@kinsi55 Thanks for posting!  We'll take a look as soon as possible.

In the mean time, there are a few ways you can help speed things along:
 - look for a workaround.  _(Even if it's just temporary, sharing your solution can save someone else a lot of time and effort.)_
 - tell us why this issue is important to you and your team.  What are you trying to accomplish?  _(Submissions with a little bit of human context tend to be easier to understand and faster to resolve.)_
 - make sure you've provided clear instructions on how to reproduce the bug from a clean install.
 - double-check that you've provided all of the requested version and dependency information.  _(Some of this info might seem irrelevant at first, like which database adapter you're using, but we ask that you include it anyway.  Oftentimes an issue is caused by a confluence of unexpected factors, and it can save everybody a ton of time to know all the details up front.)_
 - read the [code of conduct](https://sailsjs.com/documentation/contributing/code-of-conduct).
 - if appropriate, ask your business to [sponsor your issue](https://sailsjs.com/support).   _(Open source is our passion, and our core maintainers volunteer many of their nights and weekends working on Sails.  But you only get so many nights and weekends in life, and stuff gets done a lot faster when you can work on it during normal daylight hours.)_
 - let us know if you are using a 3rd party plugin; whether that's a database adapter, a non-standard view engine, or any other dependency maintained by someone other than our core team.  _(Besides the name of the 3rd party package, it helps to include the exact version you're using.  If you're unsure, check out [this list of all the core packages we maintain](https://sailsjs.com/architecture).)_ 
<hr/>

[...]

PR #6855, opened 2019-09-08
[PATCH] IN CSRF hook, only lowercase the route path if the route isn't a regex

rachaelshaw:
See https://github.com/balderdashy/sails/issues/6838


mikermcneil:
LGTM

PR #6961, opened 2020-03-13
[proposal] Support camel case for action names

shannonhochkins:
We have strict architecture patterns for our folders and filenames to be written in camelCase. 

Sails (for actions) only supports kebab-case and we'd really like the option to specify the casings in our config to be able to name our actions `findOne.js` for example.

I've tested this and it works fine and doesn't throw any errors


shannonhochkins:
Hi, has anyone managed to review this yet?

This is vital to us, the standard flow using Controllers allows camelCase, titleCase, whatever really as the action names, however splitting them up into individual actions ONLY allows kebab-case as the file names and routes.

PR #6967, opened 2020-03-29
[proposal] Add 'UsageError' handling to `parseBlueprintOptions()`

danielsharvey:
Custom `parseBlueprintOptions()` handling is configurable (see [here](https://sailsjs.com/documentation/reference/configuration/sails-config-blueprints)) 
but currently does not enable returning UsageError's  view via
`res.badRequest()` in the same way as errors that occur later
during the blueprint processing e.g. [create action](https://github.com/balderdashy/sails/blob/master/lib/hooks/blueprints/actions/create.js#L102)
Handling of UsageError's using `res.badRequest()` is added in this PR.

### DIFF
--- a/lib/hooks/blueprints/actions/add.js
+++ b/lib/hooks/blueprints/actions/add.js
@@ -23,7 +23,15 @@ module.exports = function addToCollection (req, res) {
   // Set the blueprint action for parseBlueprintOptions.
   req.options.blueprintAction = 'add';
 
-  var queryOptions = parseBlueprintOptions(req);
+  var queryOptions;
+  try {
+    queryOptions = parseBlueprintOptions(req);
+  } catch (err) {
+    switch (err.name) {
+      case 'UsageError': return res.badRequest(formatUsageError(err, req));
+      default: return res.serverError(err);
+    }
+  }
   var Model = req._sails.models[queryOptions.using];
 
   var relation = queryOptions.alias;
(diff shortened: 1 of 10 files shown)


danielsharvey:
By way of example.

Without the change a `500 Internal Server Error` error is received:
```
Internal Server Error
```

With the change a `400 Bad Request` error is received:
```
{
  "code": "E_READ_ONLY_ATTRIBUTE",
  "details": "Attribute `group`.`id` is read-only; cannot set",
  "message": "The server could not fulfill this request (`POST /api/v2/group`) due to a problem with the parameters that were sent.  See the `details` for more info."
}
```

lib/hooks/blueprints/actions/add.js
```diff
@@ -23,7 +23,15 @@ module.exports = function addToCollection (req, res) {
   // Set the blueprint action for parseBlueprintOptions.
   req.options.blueprintAction = 'add';
 
-  var queryOptions = parseBlueprintOptions(req);
+  var queryOptions;
+  try {
+    queryOptions = parseBlueprintOptions(req);
+  } catch (err) {
+    switch (err.name) {
```
sachin0903, line 30: why switch? can this be changed to
catch (err) {
if (err.name === 'UsageError') {
  return res.badRequest(formatUsageError(err, req));
}

return res.serverError(err);
}

Or
return err.name === 'UsageError' ? res.badRequest(formatUsageError(err, req)) : res.serverError(err)

This is much cleaner. I think you don't have multiple if...else so switch is not right option here.

PR #7066, opened 2020-11-09
[fixes #7063] add toJSON so pubsub obeys customToJSON

### LINKED ISSUE #7063: Removing attribute from PubSub publish
Node version: 10.6.3
Sails version (sails): 1.2.4
ORM hook version (sails-hook-orm): 2.1.1
Sockets hook version (sails-hook-sockets): 2.0.0
DB adapter & version (e.g. sails-mysql@5.55.5): sails-mysql@1.0.1
<hr/>

When use HTTP GET requests you can configure customToJSON so that you can omit specific attributes you don't want returned, e.g. password.  This is described in the docs [here](https://sailsjs.com/documentation/reference/waterline-orm/records/to-json).   For instance:

```
customToJSON: function() {
    // Return a shallow copy of this record with the password removed
    return _.omit(this, ['password'])
  },
```

My issue is that customToJSON does not seem to be obeyed when using Resouceful PubSub.  If I update my 'user' model I get something like this:

```
{
  "verb": "updated",
  "data": {
    "password": "random-new-password-obviously-not-in-plaintext",
    "id": 1
  },
  "id": 1,
  "previous": {
    "password": "old-password",
    "roles": [
      {
        "createdAt": 1603390502074,
        "updatedAt": 1603390502074,
        "id": 1,
        "name": "admin"
      }
    ],
    "createdAt": 1604509423330,
    "id": 1,
    "email": "me@gmail.ca",
    "firstName": "John",
    "lastName": "Doe"
  }
}
```

The 'previous' attribute contains the entire user object, including things that are excluded with customToJSON.
Is there some way to modify what is being published?

Tyler-RCSD:
Addresses my issue (#7063).

_publishUpdate and _publishDestroy were being passed 'previous' records in the options argument in the destroy and update blueprint actions and then being published without ever having a toJSON() called on them.  This caused the 'customToJSON' defined on the model to not be obeyed.

### DIFF
--- a/lib/hooks/blueprints/actions/destroy.js
+++ b/lib/hooks/blueprints/actions/destroy.js
@@ -66,7 +66,7 @@ module.exports = function destroyOneRecord (req, res) {
       }//-•
 
       if (req._sails.hooks.pubsub) {
-        Model._publishDestroy(criteria[Model.primaryKey], !req._sails.config.blueprints.mirror && req, {previous: record});
+        Model._publishDestroy(criteria[Model.primaryKey], !req._sails.config.blueprints.mirror && req, {previous: record.toJSON()});
         if (req.isSocket) {
           Model.unsubscribe(req, [record[Model.primaryKey]]);
           Model._retire(record);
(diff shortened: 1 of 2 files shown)


Tyler-RCSD:
This doesn't actually work.  The changes to _publishUpdate work but the changes to _publishDestroy do not.  I get an error record.toJSON is not a function.  I am not sure the reason why.

PR #7079, opened 2020-12-03
[patch] Updated "datastore configuration" link

AlanConstantino:
The datastore configuration link, when clicked on, would take you to a 404 error page.


eashaw (approved):
Thanks for catching that @AlanConstantino!

PR #7084, opened 2020-12-24
[patch] fix node 14 warnings by upgrading prompt dependency

DominusKelvin:


eashaw:
Hi @DominusKelvin, Would you be able to describe the bug this is fixing in more detail? I want to make sure I'm looking for the right thing when I test this out. Thanks for the PR!

DominusKelvin:
> Hi @DominusKelvin, Would you be able to describe the bug this is fixing in more detail? I want to make sure I'm looking for the right thing when I test this out. Thanks for the PR!

@eashaw if you run sails lift on Node.js 14 and above you get a warning about circular dependency. This patch fixes that.

DominusKelvin:
@eashaw here is a Loom video showing the reason why for this Patch
https://www.loom.com/share/afa1f768f29a4847a15b30656e1a4e5b

DominusKelvin:
@eashaw any word on this yet?

eashaw (approved):

package.json
```diff
@@ -54,7 +54,7 @@
     "parseurl": "1.3.2",
     "path-to-regexp": "1.5.3",
     "pluralize": "1.2.1",
-    "prompt": "0.2.14",
+    "prompt": "^1.1.0",
```
mikermcneil, line 57: Since `prompt` is an externally maintained dep, lets use a strict semver range `1.1.0`

mikermcneil (changes requested):

package.json
DominusKelvin: Okay that works well

PR #7108, opened 2021-03-12
Added link to deployment article

JohnGrisham:
I wrote an article detailing deployment and hosting with Azure. I think this article would be helpful to others who might struggle with these topics.


S-Stephen:
I saw this and as I'm interested in ways to deploy sails thought I'd have a look.

The first article doesn't seem to reference sailsjs at all; so would be quite confusing to have this in the official docs.
I'm unable to read the second article on medium as it's behind a paywall :-(

JohnGrisham:
> I saw this and as I'm interested in ways to deploy sails thought I'd have a look.
> 
> The first article doesn't seem to reference sailsjs at all; so would be quite confusing to have this in the official docs.
> I'm unable to read the second article on medium as it's behind a paywall :-(

Thanks for letting me know, I was also confused as to why the first article was there since it's more in line with node deployment. I'll update the link with the version that should be accessible by anyone.

JohnGrisham:
@mikermcneil Any chance this can get reviewed when a maintainer has time? Shouldn't take very long, thanks keep up the good work!

PR #7134, opened 2021-06-29
[misc] Wrong wording in update.md

anurbol:


eashaw (approved):
Thanks for catching that @anurbol!

PR #7158, opened 2021-07-27
[patch] Small text correction

zsteinkamp:
Change "under-the-hook" to "under-the-hood".


eashaw (approved):

eashaw:
@zsteinkamp Thanks for catching that!

PR #7172, opened 2021-08-25
[fixes #7168] Add support for latest version of connect-mongo

### LINKED ISSUE #7168: Latest version of connect-mongo not supported by session hook
**Node version**: v10.6.0
**Sails version** _(sails)_: v1.4.4
**ORM hook version** _(sails-hook-orm)_: v3.0.2
**Sockets hook version** _(sails-hook-sockets)_: v2.0.1
**Organics hook version** _(sails-hook-organics)_: -
**Grunt hook version** _(sails-hook-grunt)_: v4.0.1
**Uploads hook version** _(sails-hook-uploads)_: -
**DB adapter & version** _(e.g. sails-mysql@5.55.5)_: -
**Skipper adapter & version** _(e.g. skipper-s3@5.55.5)_: -
**connect-mongo version**: v4.5.0
<hr/>

The latest version of `connect-mongo` (v4) is not supported by the Sails session hook. This is a problem for me because it implements an important bug fix. 

The `connect-mongo` update is a breaking change with the key difference (mentioned in the [`connect-mongo` migration docs](https://github.com/jdesboeufs/connect-mongo/blob/master/MIGRATION_V4.md)):

> Create the store using `MongoStore.create(options)` instead of `new MongoStore(options)`

This causes an error at https://github.com/balderdashy/sails/blob/master/lib/hooks/session/index.js#L405 

```
error: Failed to lift app: Error: Encountered error attempting to instantiate a session store using the installed version of `connect-mongo` (a session adapter).
Raw error from the session adapter:
---
TypeError: Class constructor MongoStore cannot be invoked without 'new'
    at afterMaybeConnectToRedis (/opt/redbox-portal/node_modules/sails/lib/hooks/session/index.js:405:33)
    ...
```

I am happy to look at implementing a fix and submitting a PR. I assume the fix would need to be backwards compatible with the older versions of `connect-mongo`?

ElizabethForest:
I've implemented a simple fix for the connect-mongo compatibility issues - fixes #7168
The inclusion of `mongoUrl` vs. `url` was used as the flag for which version is used. I added a test to ensure sails can lift with both versions of `connect-mongo`. 

As mentioned in the issue, this is a required fix for us as older versions of `connect-mongo` didn't implement reconnection logic which was causing our app to go down and stay down regularly.


mikermcneil:
@ElizabethForest This is a great PR.  Thank you for providing your use case.

I think we should expose a way for you to have more control over this in your app.

Proposal: Minimally-viable inversion of control

1. In our [sails.config.session reference docs](https://sailsjs.com/documentation/reference/configuration/sails-config-session#?advanced-session-config), add a new config key to the table under advanced session config: `handleConstructingSessionStore`.  It'll be a function.  Its default value will be `undefined`.  Its details might say: "An optional override function for Sails to call instead of the standard session store construction behavior.  To use this setting, please first read and understand [the relevant source code](https://github.com/balderdashy/sails/blob/f399a2a2c2b8acdc2e9b9527bfa932ee747d2b19/lib/hooks/session/index.js#L405-L406)."  (Note that this is deliberately abstract to discourage folks from using this unless they feel comfortable understanding and reading the source code, because if they can't they are unlikely to be successful, and the most important thing for Sails to do is make its users successful)
2. Add a code comment in the relevant part of the session hook linking to this PR so there's context for any future changes.
3. Replace the [two lines of code in the `try`](https://github.com/balderdashy/sails/blob/f399a2a2c2b8acdc2e9b9527bfa932ee747d2b19/lib/hooks/session/index.js#L405-L406) and instead run the configured `handleConstructingSessionStore` function that was passed in from userland (aka in your config/session.js file in your Sails app).  Pass this userland function the arguments it needs to be successful (see below)
4. Right [here after the `catch` closes](https://github.com/balderdashy/sails/blob/f399a2a2c2b8acdc2e9b9527bfa932ee747d2b19/lib/hooks/session/index.js#L429), do `app.config.session.store = …;` (where `…` is the return value from the `handleConstructingSessionStore` function)
5. Bonus: Tweak the error message in the `catch` block to at least alude to the idea that `sails.config.session.handleConstructingSessionStore` could have been involved, if it was configured.





Here's an example `handleConstructingSessionStore` function you could then use in your `config/session.js` file in your Sails app:

```js
// config/session.js (in your Sails app)
//…
handleConstructingSessionStore: (sessionConfig, configuredSessionAdapter, expressSessionFromSailsCore)=>{
  // Provided for you by Sails' session hook:
  // • `sessionConfig` is what Sails has loaded for sails.config.session
  // • `configuredSessionAdapter` is the package like what you get from doing require('connect-mongo')
  // • `expressSessionFromSailsCore` is the version of express-session used by Sails core session hook, provided for your convenience
  let CustomStore = configuredSessionAdapter(expressSessionFromSailsCore);
  return new CustomStore(sessionConfig);
},
//…
```

This approach will allow folks to bring in their own version of express-session and of connect-mongo, or whatever session adapter they like, in their userland package.json file in their Sails app.


Thank you @eashaw for your help with this!  

@ElizabethForest Would you have time to implement the changes above, pulling your tests into your userland code instead of in core?  We'd be happy to merge and release this so you have a future-proof escape hatch to do what you need to have a good developer experience even if you're not using connect-redis.

ElizabethForest:
Thanks @mikermcneil! I think that will be a really nice solution :) 
I should have the time to make those changes this week/next week depending. 

Should the existing code be updated to support the latest version of `connect-mongo` by default, or left as supporting the old version?

lib/hooks/session/index.js
```diff
@@ -32,6 +32,13 @@ module.exports = function(app) {
         // By default, disable session for requests to paths that look like static assets.
         isSessionDisabled: function (req){
           return !!req.path.match(req._sails.LOOKS_LIKE_ASSET_RX);
+        },
+        // Added overrideable function for constructing the session store based on https://github.com/balderdashy/sails/pull/7172
+        // • `sessionConfig` is what Sails has loaded for sails.config.session
+        // • `configuredSessionAdapter` is the package like what you get from doing require('connect-mongo')
+        // • `expressSessionFromSailsCore` is the version of express-session used by Sails core session hook, provided for your convenience
+        handleConstructingSessionStore: (sessionConfig, configuredSessionAdapter, expressSessionFromSailsCore) => {
+          return new configuredSessionAdapter(sessionConfig);
```
ElizabethForest, line 41: At the moment this will support `connect-mongo` v4.5 I can change it to support `connect-mongo` v2 by changing the body to:
```
var CustomStore = configuredSessionAdapter(expressSessionFromSailsCore);
return new CustomStore(sessionConfig);
```

eashaw (approved):
Thank you @ElizabethForest!

PR #7174, opened 2021-09-08
Fix typo in sails-run.js

eltociear:
occured -> occurred


eashaw (approved):
Thanks for catching that @eltociear!

PR #7175, opened 2021-09-12
[proposal] Add documentation regarding a breaking change in SSL connection syntax

jarodccrowe:
The update referenced in this note is available to read here -
https://node-postgres.com/announcements#2020-02-25

Without adding the new ssl connection attribute, database connection will fail.


eashaw (approved):
Thanks @jarodccrowe!

PR #7208, opened 2022-03-03
Add documentation for select and omit

cincodenada:
select() and omit() are mentioned fleetingly in the find docs, but don't have
their own documentation. This adds basic docs for their usage.


cincodenada:
Where in the guidelines does it mention prefixing the PR title with the topic? I looked through what I think is the entire guidelines and don't think I saw it anywhere, but the sailsbot message seems to imply that I should have known about the pull request title from reading the guidelines.

PR #7210, opened 2022-03-24
[misc] Use an allowlist for all JS includes

cincodenada:
An exclusion list does not make much sense here - trying to include literally any file other than markdown and txt files - including dotfiles (if they have an extension), which results in errors lifting if there are things like editor temp files, which seems pretty unintended.

An inclusion list makes more sense here, and we have one already that we're using elsewhere, so go ahead and just use it any time we're including source files.

I ran into this because my editor left behind hidden dotfiles after editing some service files and Sails tried to require them, which was very unexpected.

By way of testing, I've added some files in the `sampleapp` that shouldn't be loaded as source. If y'all want a more specific test for this I can probably get that going, but these files successfully crash out the tests on master.


luislobo (approved):
This seems like a better approach than the original one

cincodenada:
Hello! I ran into this issue again twice in the last two days, and it materialized in two different ways, so I circled back to check on this. There were tests failing because `.ejs` files weren't included in the list, so I reworked this to add those, so tests are now passing.

PR #7263, opened 2022-11-20
Update helpers.md docs to explain subfolders

itsalaidbacklife:
Added 'Organizing helpers' section describing use of subfolders and additionally swapped all instances of `var` with `const` on the `helpers.md` page


mikermcneil:
@eashaw @itsalaidbacklife think we should set up something to autodeploy sailsjs.com just like Fleet’s?

https://m.youtube.com/watch?v=OME5lQUJ4q4&list=PLzn32d5NSCl_ZWw58HoMX-mBqJp91O1yI&index=1

I think it might just involve copying over the GitHub action into the sailsjs.com repo. It’d be great to fold everything in one place eventually and deploy out of a subdir, but I’m just thinking what might be fastest

mikermcneil (commented):
@rachaelshaw

PR #7284, opened 2023-06-25
[chore] bump up machine package to latest

DominusKelvin:


DominusKelvin:
No need to manually bump this up since the semver range will catch it when you update sails

Issue #7346, opened 2024-09-04
MariaDB support in documentation?


robertsilen:
Hello, does sails support MariaDB? I noticed that MySQL is mentioned in the [README](https://github.com/balderdashy/sails/blob/master/README.md) and e.g. on the [available adapters](https://sailsjs.com/documentation/concepts/extending-sails/adapters/available-adapters) docs page. Any reason not to add MariaDB? Its the default in most distros so only seems fair, and it eliminates any doubts about compatibility.

sailsbot:
@robertsilen Thanks for posting!  We'll take a look as soon as possible.

In the mean time, there are a few ways you can help speed things along:
 - look for a workaround.  _(Even if it's just temporary, sharing your solution can save someone else a lot of time and effort.)_
 - tell us why this issue is important to you and your team.  What are you trying to accomplish?  _(Submissions with a little bit of human context tend to be easier to understand and faster to resolve.)_
 - make sure you've provided clear instructions on how to reproduce the bug from a clean install.
 - double-check that you've provided all of the requested version and dependency information.  _(Some of this info might seem irrelevant at first, like which database adapter you're using, but we ask that you include it anyway.  Oftentimes an issue is caused by a confluence of unexpected factors, and it can save everybody a ton of time to know all the details up front.)_
 - read the [code of conduct](https://sailsjs.com/documentation/contributing/code-of-conduct).
 - if appropriate, ask your business to [sponsor your issue](https://sailsjs.com/support).   _(Open source is our passion, and our core maintainers volunteer many of their nights and weekends working on Sails.  But you only get so many nights and weekends in life, and stuff gets done a lot faster when you can work on it during normal daylight hours.)_
 - let us know if you are using a 3rd party plugin; whether that's a database adapter, a non-standard view engine, or any other dependency maintained by someone other than our core team.  _(Besides the name of the 3rd party package, it helps to include the exact version you're using.  If you're unsure, check out [this list of all the core packages we maintain](https://sailsjs.com/architecture).)_ 
<hr/>

Please remember: never post in a public forum if you believe you've found a genuine security vulnerability.  Instead, [disclose it responsibly](https://sailsjs.com/security).

For help with questions about Sails, [click here](http://sailsjs.com/support).

Maxed0utt:
[...]

PR #7352, opened 2024-10-09
add MariaDB to supported databases in README.md

robertsilen:
This PR suggests to mention MariaDB as a supported database in the README, as MariaDB is de facto supported as per discussion https://github.com/balderdashy/sails/issues/7346


ottok:
Seems the README still does not mention MariaDB at all? With [MySQL potentially in decline under Oracle's stewardship](https://optimizedbyotto.com/post/reasons-to-stop-using-mysql/), having explicit mention of MariaDB would probably be good for Sails.

